PR #24, opened 2020-09-14
feature/process-order

ssallam:


ssallam:
The provider-py image is ready in docker pull oceanprotocol/provider-py:v0.3.0

PR #37, opened 2020-12-06
Adds checkURL endpoint.

### LINKED ISSUE #16: Initialize should check if URL still works
Initialize should also check if data URL is still accessible. (maybe do a HEAD, or OPTIONS - warning, this does not work with all servers)

If not, return an error instead of transaction parameters

### LINKED ISSUE #42: Add a checkURL endpoint to check URL ContentLength and ContentType

### LINKED ISSUE #45: Add tests for datasets downloadable with ipfs protocol
Right now, all the sample datasets in the provider unit tests contain http URLs, although ipfs is also supported. We need to add extra tests to cover the possibility of files with ipfs protocol.

calina-c:
Closes #42. Closes #16. Closes #45.

### DIFF
--- a/ocean_provider/routes.py
+++ b/ocean_provider/routes.py
@@ -262,7 +262,7 @@ def checkURL():
         return jsonify(
             status="error",
             result={"contentLength":"", "contentType":""}
-        )
+        ), 400
 
     return jsonify(
         status="success",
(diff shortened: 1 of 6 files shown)


ssallam (approved):
LGTM.

mariacarmina (approved):

alexcos20 (approved):
lgtm

PR #39, opened 2020-12-09
CheckURL  created

mariacarmina:
Implemented a new endpoint called checkURL inside of routes.py file from ocean_provider package. Basically, it returns in case of success a response that includes: the status,  the content type and the content length for an uploaded file. I have used a try-except block to identify more easily when an error occurs and I have also integrated the wrapper class Response for the return statement.

### DIFF
--- a/ocean_provider/routes.py
+++ b/ocean_provider/routes.py
@@ -287,25 +287,27 @@ def checkURL(url: string):
     Content length of a file sent through an URL.
 
     """
-    r = requests.get(url)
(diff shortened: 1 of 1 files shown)


calina-c:
@mariacarmina @alexcos20 this is replaced by #37. Can we close it?

mariacarmina:
> @mariacarmina @alexcos20 this is replaced by #37. Can we close it?

I don't mind, so...

PR #43, opened 2021-01-05
[WIP] Adds URL checking to initialize.

### LINKED ISSUE #16: Initialize should check if URL still works

calina-c:
Closes #16. Marked as draft PR and WIP for testing and further guidance.

### DIFF
--- a/ocean_provider/routes.py
+++ b/ocean_provider/routes.py
@@ -3,6 +3,7 @@
 import json
 import logging
 import os
+import requests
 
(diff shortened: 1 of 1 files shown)


alexcos20:
It shoud pay off to use the same function  (get_url_details) by this PR and also by #37.    After that, it's easier to maintain

PR #44, opened 2021-01-08
Issue#36

mariacarmina:
Added a new dictionary ('serviceEndpoints') in run.py file, in function version with all types of existed endpoints from routes.py. Tested the functionality by implementing a new unit test in test_compute.py.


ocean_provider/routes.py
```diff
@@ -86,7 +86,6 @@ def simple_flow_consume():
         'transferTxId'
     ]
     data = get_request_data(request)
-
```
calina-c, line 89: Please add this line back. Files not affected by commit should not be changed, and the empty line serves for code readability and delimitation.

ocean_provider/run.py
```diff
@@ -25,11 +25,27 @@ def get_version():
 
 @app.route("/")
 def version():
+    """
```
calina-c, line 28: This doc comment should describe what the entire function is doing, not just the changes you made. Please use a compliant description if you choose to add one.

I would suggest something like:
Displays data necessary for using the provider, including software, version, network url, provider address and a list of services endpoints that may be used.

ocean_provider/run.py
```diff
@@ -25,11 +25,27 @@ def get_version():
...
+    to store the services' routes.
+    It is open for modifications, by simply adding the service
+    name and the url for it.    """
+
     info = dict()
     info['software'] = Metadata.TITLE
     info['version'] = get_version()
     info['network-url'] = config.network_url
     info['provider-address'] = get_provider_wallet().address
+    # Added the expose endpoint. Here are the URLs to the
```
calina-c, line 39: I think this comment is unnecessary and also relates more to your changes than the general intent of the code. If the tests change, nobody will know that this is referenced here and it will result in inconsistencies with the code.

The code is very good and self-descriptive, it does not need commenting. But if you want to leave a comment, you can say something about the intention, e.g. "# URLs to the existing endpoints in services"

tests/test_compute.py
```diff
@@ -15,13 +15,31 @@
 from tests.test_helpers import (
     get_consumer_wallet,
     get_publisher_wallet,
-    get_dataset_ddo_with_compute_service_no_rawalgo, get_dataset_ddo_with_compute_service_specific_algo_dids, get_algorithm_ddo,
-    get_dataset_ddo_with_compute_service, get_compute_job_info, get_possible_compute_job_status_text, mint_tokens_and_wait, get_nonce,
+    get_dataset_ddo_with_compute_service_no_rawalgo, get_dataset_ddo_with_compute_service_specific_algo_dids,
```
calina-c, line 18: did you run isort on these imports? They do not seem to be ordered alphabetically, and actually it seems you haven't imported anything new, so no need to change the imports section.

tests/test_compute.py
```diff
@@ -15,13 +15,31 @@
...
+    assert 'software' in result
+    assert 'version' in result
+    assert 'network-url' in result
+    assert 'provider-address' in result
+    assert res.status == '200 OK'
+    # The length of the result['servicesEndpoints']
+    # is hardcoded by now. If we will add another
+    # services, we need to increase the value of the
+    # length.
+    assert len(result['servicesEndpoints']) == 5
```
calina-c, line 40: 👍

tests/test_compute.py
```diff
@@ -78,7 +96,7 @@ def test_compute_norawalgo_allowed(client):
         data=json.dumps(payload),
         content_type='application/json'
     )
-    assert response.status == '400 BAD REQUEST', f'start compute job failed: {response.status} , { response.data}'
+    assert response.status == '400 BAD REQUEST', f'start compute job failed: {response.status} , {response.data}'
```
calina-c, line 99: there are also some changes in this file that are not related to the task. Are the changes intended?

calina-c (changes requested):

mariacarmina (commented):
I have modified the doc and some indentations/spaces.

ocean_provider/run.py
```diff
@@ -25,11 +25,31 @@ def get_version():
...
+    # existed endpoints. Check test_compute.py, in function
+    # test_compute_expose_endpoint.
+    info['servicesEndpoints'] = {
+        "access": url_for('services.download'),
+        "compute": url_for('services.compute_get_status_job'),
+        "nonce": url_for('services.get_user_nonce'),
+        "encrypt": url_for('services.encrypt'),
+        "initialize": url_for('services.initialize'),
+    }
```
ssallam, line 52: does this include all the endpoints?

ssallam: missing: 
compute_start_job, compute_stop_job and compute_delete_job

ssallam (changes requested):

ocean_provider/run.py
mariacarmina: Ok, thanks. I will modify that issue too.

PR #46, opened 2021-01-08
Expose service endpoints

### LINKED ISSUE #36: Endpoint expose
On the root endpoint, each provider should expose a json that looks like:

{
   "network-url":"mainnet",
   "provider-address":"0xeAFDC69612a8bF720FBfE6A5520Cfede69a9a5b5",
   "software":"Provider",
   "version":"0.4.0",
   "serviceEndpoints":{
      "download": ("GET", "/api/v1/services/download"),
      "computeStart": ("POST", "/api/v1/services/compute"),
      "computeDelete": ("DELETE", "/api/v1/services/compute"),
      "computeStop": ("PUT", "/api/v1/services/compute"),
      "computeStatus": ("GET", "/api/v1/services/compute"),
      "nonce": ("GET", "/api/v1/services/nonce"),
      "encrypt": ("POST", "/api/v1/services/encrypt"),
      "initialize": ("GET", "/api/v1/services/initialize"),
   }
}

this will allow dynamic services types

The provider already has the root endpoint. This task adds the `serviceEndpoints` dict to that.

* Include all the provider service endpoints (preferrably using a dynamic approach)
* Include the HTTP method along with the endpoint url
* Remove/delete the simple flow endpoint (`simple_flow_consume`)

mariacarmina:
Added a new dictionary ('serviceEndpoints') in run.py file, in function version with all types of existed endpoints from routes.py. Tested the functionality by implementing a new unit test in test_compute.py.

closes #36

### DIFF
--- a/ocean_provider/routes.py
+++ b/ocean_provider/routes.py
@@ -379,7 +379,7 @@ def download():
 
 
 @services.route('/compute', methods=['DELETE'])
-def compute_delete_job():
+def computeDelete():
     """Deletes a workflow.
 
     ---
@@ -444,7 +444,7 @@ def compute_delete_job():
 
 
 @services.route('/compute', methods=['PUT'])
-def compute_stop_job():
+def computeStop():
     """Stop the execution of a workflow.
 
     ---
@@ -513,7 +513,7 @@ def compute_stop_job():
 
 
 @services.route('/compute', methods=['GET'])
-def compute():
+def computeStatus():
     """Get status for a specific jobId/documentId/owner
 
     ---
@@ -604,7 +604,7 @@ def compute():
 
 
 @services.route('/compute', methods=['POST'])
-def compute_start_job():
+def computeStart():
     """Call the execution of a workflow.
 
     ---
--- a/ocean_provider/routes.py
+++ b/ocean_provider/routes.py
@@ -57,7 +57,7 @@ logger = logging.getLogger(__name__)
 
 
 @services.route('/nonce', methods=['GET'])
-def get_user_nonce():
+def nonce():
     required_attributes = [
         'userAddress',
     ]
@@ -280,7 +280,7 @@ def initialize():
 
 
 @services.route('/download', methods=['GET'])
-def download():
+def access():
     """Allows download of asset data file.
 
     ---
@@ -513,7 +513,7 @@ def compute_stop_job():
 
 
 @services.route('/compute', methods=['GET'])
-def compute_get_status_job():
+def compute():
     """Get status for a specific jobId/documentId/owner
 
     ---
--- a/ocean_provider/routes.py
+++ b/ocean_provider/routes.py
@@ -280,7 +280,7 @@ def initialize():
 
 
 @services.route('/download', methods=['GET'])
-def access():
+def download():
     """Allows download of asset data file.
 
     ---
--- a/ocean_provider/run.py
+++ b/ocean_provider/run.py
@@ -19,6 +19,17 @@ provider_url = config.get(ConfigSections.RESOURCES, 'ocean_provider.url')
 blocked_url = ['services.simple_flow_consume']
 
 
+def get_services_endpoints():
+    services_endpoints = dict(map(lambda url: (url.endpoint.replace('services.', ''), url),
+                                  filter(lambda url: url.endpoint.startswith('services.')
+                                                     and url.endpoint not in blocked_url,
+                                         app.url_map.iter_rules())))
+    for (key, value) in services_endpoints.items():
+        services_endpoints[key] = (list(map(str, filter(lambda method: str(method) not in ["OPTIONS", "HEAD"],
+                                                        value.methods)))[0], str(value))
+    return services_endpoints
+
+
 def get_version():
     conf = configparser.ConfigParser()
     conf.read('.bumpversion.cfg')
@@ -34,8 +45,8 @@ def version():
         - network url;
         - provider address;
         - services endpoints, which has all
-        the existing endpoints from routes.py with
-        GET method only which are not in blocked_url.
+        the existing endpoints from routes.py
+        which are not in blocked_url.
     """
 
     info = dict()
@@ -43,11 +54,7 @@ def version():
     info['version'] = get_version()
     info['network-url'] = config.network_url
     info['provider-address'] = get_provider_wallet().address
-    info['servicesEndpoints'] = dict(map(lambda url: (url.endpoint.replace('services.', ''), '%s' % url),
-                                         filter(lambda url: url.endpoint.startswith('services.')
-                                                            and 'GET' in url.methods
-                                                            and url.endpoint not in blocked_url,
-                                                app.url_map.iter_rules())))
+    info['servicesEndpoints'] = get_services_endpoints()
     return jsonify(info)
 
 
--- a/ocean_provider/run.py
+++ b/ocean_provider/run.py
@@ -3,7 +3,7 @@
 
 import configparser
 
-from flask import jsonify, url_for
+from flask import jsonify
 from flask_swagger import swagger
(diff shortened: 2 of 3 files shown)


calina-c (approved):
Great job, @mariacarmina!

ocean_provider/run.py
```diff
@@ -25,11 +25,30 @@ def get_version():
...
+        - provider address;
+        - services endpoints, which has all
+        the existing endpoints from route.py.
+    """
     info = dict()
     info['software'] = Metadata.TITLE
     info['version'] = get_version()
     info['network-url'] = config.network_url
     info['provider-address'] = get_provider_wallet().address
+    info['servicesEndpoints'] = {
```
calina-c, line 42: @ssallam @alexcos20 I was wondering about the logic here. If we want these to be constantly filled with all services routes, without manual adding every time, we can do that. However, it seems like the services are customly named: e.g. 'access' instead of 'download' etc.

Are you ok with manual adding to this dictionary, whenever new services are added, or should we work on something that automatically takes all the services available? E.g.:

`{url.endpoint.replace('services.', ''): '%s' % url for url in app.url_map.iter_rules() if url.endpoint.startswith('services.')}`

mariacarmina: I also had this question and I understood that I can create a dictionary for manual adding the services, but if there are more suggestions about this, please let me know and I will refactor it. @ssallam & @alexcos20 .

calina-c: There is always the option of having a naming override dictionary. Consider all keys equivalent to the route names unless overridden. E.g.
Screenshot 2021-01-09 at 12 23 44

ocean_provider/run.py
```diff
@@ -25,11 +25,30 @@ def get_version():
...
+        'access': url_for('services.download'),
+        'compute': url_for('services.compute_get_status_job'),
```
ssallam, line 44: maybe call this `job_status`

ocean_provider/run.py
ssallam: Automatically generating the service endpoints dict makes sense and having an override for specific names would be great. I suggest renaming the functions if necessary so that the names don't need override.

ocean_provider/run.py
ssallam: if we generate the dict dynamically, then we can rename the function names to reflect what we want in the dict keys.

mariacarmina: At my first PR, I kept just the job_status function, because all functions related to 'compute' service have the exact same route and that's why I didn't include them. Now, I saw that it's better to have a dynamic dict, so I will refactor this.

mariacarmina:
I uploaded a new implementation. Tell me if I need to include other endpoints too.

mariacarmina:
Note: If the code alignment / indent is not correct, let me know, because PyCharm reformats the file as it wants. :(

ocean_provider/routes.py
```diff
@@ -343,7 +343,7 @@ def initialize():
 
 
 @services.route('/download', methods=['GET'])
-def download():
+def access():
```
kremalicious, line 346: might have missed some discussion for this change but download !== access. Giving access or what we usually call `initialize` happens for both cases, download & compute. So it seems more logical to keep this name as `download`

mariacarmina: I asked @alexcos20 about the task and he told me that the access service is related to download route so... That's why I modified and @ssallam told me to change the functions' names in some previous commets here. Let me know what should I modify.

ssallam: I would keep the `download` name as it directly describes what it does.

mariacarmina: So then should I keep 'access' or download is enough?

kremalicious: well, we found a conceptual flaw because of legacy. In this context it really makes sense to keep the name as `download` but since the beginning the service type in the DDO is called access service, hence the confusion. It should be named `download` too but I don't think we'll do that change

alexcos20: Agreed, keep it as download

tests/test_compute.py
```diff
@@ -31,6 +31,18 @@
...
+def test_compute_expose_endpoints(client):
+    get_response = client.get('/')
+    result = get_response.get_json()
+    assert 'servicesEndpoints' in result
+    assert 'software' in result
+    assert 'version' in result
+    assert 'network-url' in result
+    assert 'provider-address' in result
+    assert get_response.status == '200 OK'
+    assert len(result['servicesEndpoints']) == 4
```
alexcos20, line 43: maybe check for each service that we know we have in this version of provider..  it's longer, but safer

ocean_provider/routes.py
mariacarmina: Ok, thank you for the information. I will keep 'download'

tests/test_compute.py
mariacarmina: Ok, I will do that. Thank you for feedback.

alexcos20:
I think that we should add more on this.  I only see:

{"network-url":"http://172.15.0.3:8545","provider-address":"0xe08A1dAe983BC701D05E492DB80e0144f8f4b909",
"servicesEndpoints": 
 {
"compute":"/api/v1/services/compute",
"download":"/api/v1/services/download",
"initialize":"/api/v1/services/initialize",
"nonce":"/api/v1/services/nonce"

},"software":"Provider","version":"0.4.1"}

what about encryptUrl ?

mariacarmina:
> I think that we should add more on this. I only see:
> 
> {"network-url":"[http://172.15.0.3:8545","provider-address":"0xe08A1dAe983BC701D05E492DB80e0144f8f4b909](http://172.15.0.3:8545%22,%22provider-address%22:%220xe08A1dAe983BC701D05E492DB80e0144f8f4b909)",
> "servicesEndpoints":
> {
> "compute":"/api/v1/services/compute",
> "download":"/api/v1/services/download",
> "initialize":"/api/v1/services/initialize",
> "nonce":"/api/v1/services/nonce"
> 
> },"software":"Provider","version":"0.4.1"}
> 
> what about encryptUrl ?

I filtered services by 'GET' method and encrypt is basically a 'POST'. I will introduce that, too.

tests/test_compute.py
```diff
@@ -23,14 +25,32 @@
...
+    get_response = client.get('/')
+    result = get_response.get_json()
+    assert 'servicesEndpoints' in result
+    assert 'software' in result
+    assert 'version' in result
+    assert 'network-url' in result
+    assert 'provider-address' in result
+    assert get_response.status == '200 OK'
+    assert len(result['servicesEndpoints']) == len(
+        dict(map(lambda url: (url.endpoint.replace('services.', ''), '%s' % url),
```
ssallam, line 47: generating this dict can go into its own function, then can be called here in the endpoint above.

tests/test_compute.py
```diff
@@ -23,14 +25,32 @@
...
+                 filter(lambda url: url.endpoint.startswith('services.')
+                                    and 'GET' in url.methods
```
ssallam, line 49: Why include on the endpoints with the `GET` method?

ocean_provider/run.py
```diff
@@ -25,11 +38,23 @@ def get_version():
...
+        the existing endpoints from routes.py
+        which are not in blocked_url.
+    """
+
     info = dict()
     info['software'] = Metadata.TITLE
     info['version'] = get_version()
     info['network-url'] = config.network_url
     info['provider-address'] = get_provider_wallet().address
+    info['servicesEndpoints'] = get_services_endpoints()
```
ssallam, line 57: one request, can you please rename the key `servicesEndpoints` to `serviceEndpoints`

ssallam (approved):
LGTM.

PR #48, opened 2021-01-11
[Proposal] Adds flake8 and git hook.

calina-c:
Another possible improvement to the deployment flow: implement a git pre-commit hook calling flake8 to avoid formatting issues on commit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,14 +8,14 @@ This is feature complete and is a BETA version.
 ## Starting the server locally
 
 ### Quick start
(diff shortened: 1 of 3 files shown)


kremalicious (approved):
yay, automating best practices, insta-approve!

But I'm no python person so only approving out of principle :-) Maybe @ssallam wants to have another look here. If this is agreed upon I would suggest adding this at least also to ocean.py

calina-c:
By the way, since most conflicts are in the imports, running isort on all commits should make it easier to spot differences when solving conflicts, too, since both versions will be ordered and the conflicts minimal.

alexcos20 (approved):

PR #57, opened 2021-01-13
Add endpoint for assetInfo

### LINKED ISSUE #52: Provider: new endpoint assetInfo
For a given DID, the endpoint should return an array with info about the files contained by the asset

[
{ index: 0, valid: true , {"contentLength": 100, "contentType": application/text} },
{ index: 1, valid: true , {"contentLength": 200, "contentType": application/text} }
{ index: 2, valid: true , {"contentLength": 300, "contentType": application/text} }
]

ssallam:
closes #52

### DIFF
--- a/ocean_provider/routes.py
+++ b/ocean_provider/routes.py
@@ -261,13 +261,15 @@ def checkURL():
 
     if not valid:
         return jsonify(
-            status="error",
-            result={"contentLength":"", "contentType":""}
+            valid=False,
+            contentLength="",
+            contentType=""
         ), 400
 
     return jsonify(
-        status="success",
-        result=details
+        valid=True,
+        contentLength=details['contentLength'],
+        contentType=details['contentType']
     )
 
 
--- a/ocean_provider/routes.py
+++ b/ocean_provider/routes.py
@@ -213,8 +213,8 @@ def encrypt():
         return jsonify(error=str(e)), 500
 
 
-@services.route('/checkURL', methods=['POST'])
-def checkURL():
+@services.route('/fileinfo', methods=['GET'])
+def fileInfo():
     """Retrieves Content-Type and Content-Length from the given URL.
 
     This can be used by the publisher of an asset to check basic information
@@ -264,8 +264,8 @@ def checkURL():
     )
 
 
-@services.route('/assetInfo', methods=['POST'])
-def assetInfo():
+@services.route('/fileinfo/<did>', methods=['GET'])
+def assetFilesInfo(did: str):
     """Return info about each file in the asset (index, valid, contentLength, contentType)
 
     This can be used by the publisher of an asset to check basic information
@@ -273,6 +273,8 @@ def assetInfo():
     and Content-Length of the request, using primarily OPTIONS, with fallback
     to GET. In the future, we will add a hash to make sure that the file was
     not tampered with at consumption time.
+
+    :param did: str the id of the asset
 
     tags:
       - services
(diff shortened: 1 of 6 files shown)


ocean_provider/routes.py
```diff
@@ -273,6 +273,56 @@ def checkURL():
...
+    )
+    if msg:
+        return jsonify(error=msg), status
+
+    did = data['documentId']
+    asset = get_asset_from_metadatastore(get_metadata_url(), did)
+    url_list = get_asset_urls(asset, provider_wallet)
+    files_info = []
+    for i, url in enumerate(url_list):
+        download_url = get_download_url(url, app.config['CONFIG_FILE'])
```
calina-c, line 313: I'm not sure, but I believe if get_asset_urls is called before, we don't need to convert again with get_download_url. Maybe that's why the ipfs test is failing.  I'll check again tomorrow.

calina-c:
@ssallam the ipfs test failed because when running OPTIONS on the URL (which was https://gateway.ipfs.io/ipfs/QmQfpdcMWnLTXKKW9GPV7NgtEugghgD6HgzSF6gSrp2mL9, btw) didn't get the content length and type, while the GET request did. This was part of my checkURL functionality, which is why I corrected it in place. I hope I didn't overstep.

 I also changed the text/csv type for one of the existing assets, please check.

ocean_provider/routes.py
ssallam: correct, thanks for the change.

ssallam:
Lookds good, this is ready to merge.

ocean_provider/routes.py
```diff
@@ -273,6 +264,55 @@ def checkURL():
     )
 
 
+@services.route('/assetInfo', methods=['POST'])
```
kremalicious, line 267: using uppercase/lowercase in HTTP paths could really go wrong. I would suggest always using lowercase when defining API paths.

Also, what's the difference between `/checkURL` & `/assetInfo` and why are they separate?

alexcos20: checkUrl takes an URL (http:///www.example.com) and check if your selected provider is able to access it and gets file information (length & type)
/assetInfo takes a DID with some files (you don't have the URLs) and will send back the same interface, but with updated values.

Let's say that you published www.example.com as DID1. and you are updating the content, but in the DDO you still have the old length of the file. By calling /assetInfo, provider can decrypt the URLs , check if they are still valid, get the file info and then you can update your DDO with the new info

kremalicious: Cool, got it. So if both return the same thing maybe the paths should more reflect what they do and be a bit grouped together:

- `/check-url`
- `/check-url-encrypted`

But what we are actually checking and returning is info about the file behind the URL. So named after what they return, and more REST like maybe:

- `/fileinfo`
- `/fileinfo/encrypted`

ssallam: the fileinfo endpoint seems more logical, but I would drop the `decrypt` part.

kremalicious: I mean even better if both endpoints could be combined into one `/fileinfo` endpoint and depending if passed value is a URI or a DID different functionality happens but with the same result

ssallam: now thinking this may look better / easier to understand:
- `GET /fileinfo/<url>`
- `GET /assetfilesinfo/<did>`

kremalicious: if we want to make it future-proof and support multiple files then `/fileinfo` should in both cases return array of objects, then there's also no further need to split it up

alexcos20: Update:  let's have POST /fileinfo  with a payload { "url":"xx"}  or { "did":"xx"}

calina-c: I'm on it.

alexcos20:
hold it for minute...
if we have a DID with 3 files, of which 
index 0 -  OK
index 1 - NOK
index 2 - OK  
what is the result?

if we end up with an array of 2 files, it's not good.

alexcos20 (approved):
lgtm

PR #58, opened 2021-01-18
Adds access token endpoint, model and unit test to retrieve. Improve request validation.

### LINKED ISSUE #55: Provider: Auth tokens
Provider can generate an access token for a consumer , for a specified asset, for a specific expiry time if the user can provide the txid of the ownership (very similar with the download endpoint, but instead of download, it will generate the access_token)

You cannot request more that 1 access_tokens for the same txId with multiple consumers  (it's fine to request multiple access_tokens for the same txId and the same consumer)

Provider will keep the token into a database, along with used=0  , expiry = timestamp of expiration date, did, address, etc

At usage, used is incremented, so that auth token cannot be used anymore

Access token = did + consumer , signed by the provider with the provider's private key

On consume, Provider will request access token  (message) + signature  of that message signed by the consumer and compare it with the database

For compute 2 data , the consumer will request access token for each dataset (algo, etc) , it's going to sign the requests using the auth tokens and provider will generate access urls that are going to be passed from the provider to operator - service , all the way till pod-configure

This access_token can be used for download endpoint instead of nonce

calina-c:
Fixes #55.

Changes proposed in this PR:

- add an `access-token` endpoint to generate an access token for compute to data
- optionally use that access token when downloading the file, instead of direct computation.
- improves request validation by adding Flask Sieve, request classes and validation decorators

Pending as WIP:
- [x] actual token building (uuid4 id now a placeholder) -> replaced with sha256(uuid4())
- [x] actual token expiration time handling (15 minutes from now is a placeholder) -> parameter from libs, no. of seconds
- [x] _which parameters from download endpoint will become "obsolete" and can be replaced with the access_token_ -> download can function with only access_token & did (+ fileIndex, serviceId), in another issue TODO.
- [x] should I also add the service type in the access_token table?
- [x] for the validation to have one single consumer per txId, should I consider the existence of the token itself, or should I disregard expired tokens? -> add txId in datbase too, does not matter if expired

Question:
- do we rely on the 401 code for invalid signature anywhere, or is 400 ok, as part of the bad request handling? Do I need to add special handling to return 401 when signature is invalid?


ssallam:
Great work on improving the validator.

ssallam:
> do we rely on the 401 code for invalid signature anywhere, or is 400 ok, as part of the bad request handling? Do I need to add special handling to return 401 when signature is invalid?
It's ok as long as the error message is clear.

calina-c:
> > do we rely on the 401 code for invalid signature anywhere, or is 400 ok, as part of the bad request handling? Do I need to add special handling to return 401 when signature is invalid?
> 
> It's ok as long as the error message is clear.

Yes, sieve comes with default validation messages, and I've added the custom ones for signature and access_token issues. I think validation is much more clear to the caller now, since all validation failures are returned.

ocean_provider/requests.py
```diff
@@ -0,0 +1,181 @@
...
+        self._processor = CustomRulesProcessor()
+
+
+class CustomRulesProcessor(RulesProcessor):
+    def validate_signature(self, value, params, **kwargs):
+        self._assert_params_size(size=3, params=params, rule='signature')
+        owner = self._attribute_value(params[0])
+        did = self._attribute_value(params[1])
+        rule = params[2]
+        original_msg = f'{owner}{did}' if rule == 'customer_did' else f'{did}'
```
ssallam, line 46: what is `customer_did`?

ocean_provider/requests.py
```diff
@@ -0,0 +1,181 @@
...
+        return False
+
+    def validate_access_token(self, value, params, **kwargs):
+        self._assert_params_size(size=2, params=params, rule='access_token')
+        did = self._attribute_value(params[0])
+        if did.startswith('did:'):
+            did = add_0x_prefix(did_to_id(did))
+
+        consumer_address = self._attribute_value(params[1])
+        tx_id = value
```
ssallam, line 64: is `value` an `access_token` or a tx_id?

ocean_provider/routes.py
```diff
@@ -6,27 +6,31 @@
...
 from ocean_utils.agreements.service_types import ServiceTypes
 from ocean_utils.did import did_to_id
 from ocean_utils.http_requests.requests_session import get_requests_session
 
+from ocean_provider.access_token import AccessToken
 from ocean_provider.exceptions import BadRequestError, InvalidSignatureError
 from ocean_provider.log import setup_logging
 from ocean_provider.myapp import app
+from ocean_provider.requests import (AccessTokenRequest, ComputeRequest,
+                                     ComputeStartRequest, DownloadRequest,
```
ssallam, line 22: can we change the style of imports by starting on new line and keeping only one import per line?

ocean_provider/routes.py
```diff
@@ -439,25 +403,108 @@ def download():
...
+            f"serviceId={data.get('serviceId')}"
+            f"serviceType={data.get('serviceType')}",
+            exc_info=1
+        )
+        return jsonify(error=str(e)), 500
+
+
+@services.route('/access-token', methods=['GET'])
+@validate(AccessTokenRequest)
+def access_token():
```
ssallam, line 420: since the function name will be visible in the exposed endpoints, and other functions use camel case, let's stick with that naming here.

ocean_provider/requests.py
calina-c: the rule for the original message to check in the signature: it is either customer_address + did OR just the did. This is a string defining the customer_address + did rule.

ocean_provider/routes.py
```diff
@@ -439,25 +403,108 @@ def download():
...
+            f"Payload was: documentId={data.get('did')}, "
+            f"consumerAddress={data.get('consumerAddress')},"
+            f"serviceId={data.get('serviceId')}"
+            f"serviceType={data.get('serviceType')}",
+            exc_info=1
+        )
+        return jsonify(error=str(e)), 500
+
+
+@services.route('/access-token', methods=['GET'])
```
ssallam, line 418: rename to `/accesstoken`

ocean_provider/requests.py
calina-c: a tx_id, since the validator is set on the txId field.

ocean_provider/requests.py
ssallam: i think we use `consumer` typically

ocean_provider/routes.py
calina-c: done ✅

ocean_provider/requests.py
ssallam: where is the access token?

ocean_provider/routes.py
ssallam: isort or black, are they configureable?

ocean_provider/routes.py
ssallam: this is not the endpoint, it is just the function name, which is used in the exposed endpoints dict when invoking the root endpoint

ocean_provider/requests.py
calina-c: done ✅

ocean_provider/requests.py
calina-c: we check that there is no access_token for these values (txId, consumerAddress and did). If there is one in the database, then we reject the request with 400.

ocean_provider/routes.py
calina-c: black not very much, but it is the standard in formatting. I could configure isort perhaps, but we would still end up with this style after converting to black.

ocean_provider/routes.py
calina-c: done ✅

ssallam (approved):
LGTM.

calina-c:
Closing since it was superseded by the payer/owner functionality. If we ever resurrect this, it is not difficult to do, but either way we won't adapt this branch, only use it as inspiration.

PR #60, opened 2021-01-20
Adds access token usage in download.

calina-c:
Related to #55 .

Changes proposed in this PR:

- adds token usage in download endpoint

### DIFF
--- a/ocean_provider/access_token.py
+++ b/ocean_provider/access_token.py
@@ -125,11 +125,9 @@ class AccessTokenStorage(StorageBase):
                 )
             ]
 
-            (
-                consumer_address, access_token
-            ) = rows[0] if rows else (None, None)
+            consumer_address, access_token = rows[0] if rows else (None, None)
 
-            return (consumer_address, access_token)
+            return consumer_address, access_token
 
         except Exception as e:
             logging.error(
--- a/ocean_provider/access_token.py
+++ b/ocean_provider/access_token.py
@@ -30,9 +30,8 @@ class AccessToken:
             did, consumer_address, tx_id, delegate_address
         )
 
-    def use_access_token(self, address):
-        # TODO
-        pass
+    def get_access_token(self, delegate_address, did, tx_id):
+        return self.storage.get_access_token(delegate_address, did, tx_id)
 
 
 class AccessTokenStorage(StorageBase):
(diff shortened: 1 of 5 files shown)


ocean_provider/utils/encryption.py
```diff
@@ -1,15 +1,29 @@
 import ecies
+from ocean_lib.web3_internal.wallet import Wallet
 from web3 import Web3
 
 from ocean_provider.utils.accounts import get_private_key
 
 
 def do_encrypt(document, provider_wallet):
```
ssallam, line 8: using 2 params instead of the overloaded use of `provider_wallet` makes it more clear.

tests/test_access_token.py
```diff
@@ -86,14 +89,15 @@ def test_access_token(client):
...
+    payload.pop('delegatePublicKey')
     request_url = at_endpoint + '?' + '&'.join(
         [f'{k}={v}' for k, v in payload.items()]
     )
     response = client.get(request_url)
+    # missing delegatePublicKey
     assert response.status_code == 400, f'{response.data}'
 
-    payload['delegateAddress'] = some_wallet.address
+    payload['delegatePublicKey'] = get_private_key(some_wallet).public_key
```
ssallam, line 100: please make sure public_key is a string not an object, or at least that it resolves to str when dumped to json.

ocean_provider/access_token.py
```diff
@@ -111,3 +110,29 @@ def check_unique(self, did, consumer_address, tx_id, delegate_address):
...
+                       AND expiry_time > ?;''',
+                    (delegate_address, did, tx_id, datetime.now())
+                )
+            ]
+
+            (
+                consumer_address, access_token
+            ) = rows[0] if rows else (None, None)
+
+            return (consumer_address, access_token)
```
ssallam, line 132: ( ) are not necessary

ocean_provider/requests.py
```diff
@@ -56,28 +56,44 @@ def validate_signature(self, value, params, **kwargs):
 
         return False
 
+    def validate_download_signature(self, value, params, **kwargs):
+        self._assert_params_size(size=3, params=params, rule='signature')
+        owner = self._attribute_value(params[0])
```
ssallam, line 61: is owner here the delegate? if so delegate_address makes it more clear

tests/test_access_token.py
calina-c: '&'.join([f'{k}={v}' for k, v in payload.items()]) will perform an automatic string conversion on this object, and is indeed what I recieve through the payload. Do you want me to add an extra conversion step?

ssallam: perfect.

ssallam (approved):
Great.

PR #63, opened 2021-01-26
create new sqlite db connection on each call

### LINKED ISSUE #62: nonce update fails due to sqlite thread error
Problem:
Updating the nonce in the sqlite db fails when using threaded worker type

Solution:
Create a  new db connection on each update or within the endpoint handler

ssallam:
Fixes #62

Changes proposed in this PR:

- make a new db connection on each call

### DIFF
--- a/ocean_provider/user_nonce.py
+++ b/ocean_provider/user_nonce.py
@@ -10,7 +10,11 @@ class UserNonce:
 
     def __init__(self, storage_path=None):
         self._storage_path = storage_path
-        self.storage = NonceStorage(storage_path)
+        NonceStorage(True, storage_path)
+
+    @property
+    def storage(self):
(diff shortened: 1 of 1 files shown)


ocean_provider/user_nonce.py
```diff
@@ -10,7 +10,11 @@ class UserNonce:
 
     def __init__(self, storage_path=None):
         self._storage_path = storage_path
-        self.storage = NonceStorage(storage_path)
+        NonceStorage(True, storage_path)
```
calina-c, line 13: can you please name this parameter explicitly? e.g. NonceStorage(storage_path, create_table=True)

ssallam: done

calina-c (approved):

PR #69, opened 2021-01-27
Request improvements

calina-c:
Changes proposed in this PR:

- add request validation with Flask Sieve
- add SQL Alchemy and use in conjunction to Nonce
- (this was salvaged from the access token branch, where improvements were combined with the AT functionality, which is now in doubt).


ocean_provider/myapp.py
```diff
@@ -2,14 +2,42 @@
...
+PROJECT_ROOT = dirname(dirname(abspath(__file__)))
+app.config['SQLALCHEMY_DATABASE_URI'] = 'sqlite:////' + os.path.join(
+    PROJECT_ROOT,
+    get_config().storage_path
+)
+
+db = SQLAlchemy(app)
+
+
+class UserNonce(db.Model):
```
ssallam, line 33: why is this in `myapp.py`

ocean_provider/user_nonce.py
```diff
@@ -1,79 +1,29 @@
...
-        if nonce is not None:
-            return int(nonce)
-        return UserNonce.FIRST_NONCE
+def increment_nonce(address):
+    nonce_object = UserNonce.query.filter_by(address=address).first()
+    if nonce_object:
+        nonce_value = nonce_object.nonce
+    else:
+        nonce_object = UserNonce(address=address)
+        nonce_value = UserNonce.FIRST_NONCE
```
ssallam, line 20: can it be assumed that nonce object exists for this address, otherwise why do increment_nonce

ocean_provider/myapp.py
calina-c: it needs too be before db.create_all(), so that it is recognized. I'll try to find other solutions.

ocean_provider/user_nonce.py
calina-c: I think so, but I would rather be on the safe side...

ocean_provider/myapp.py
calina-c: @ssallam I moved them to a models file.

ssallam (approved):
LGTM.

trentmc:
@ssallam to merge

PR #74, opened 2021-01-27
secure urls

alexcos20:
Secures urls

Changes proposed in this PR:
  
Any URL must be verified
- if it's a valid URL
- if the domain resolves to an ip address
- if the ipaddress is a public one (not loopback, not reserved, etc)

### DIFF
--- a/ocean_provider/util_url.py
+++ b/ocean_provider/util_url.py
@@ -84,6 +84,7 @@ def validate_dns_record(record, domain, record_type):
                     f"[!] DNS record type {record_type} for domain name "
                     f"{domain} resolves to a non public IP address {value}, "
                     "but allowed by config!"
+                    return True
                 )
             else:
                 logger.error(
--- a/ocean_provider/util_url.py
+++ b/ocean_provider/util_url.py
@@ -11,7 +11,9 @@ logger = logging.getLogger(__name__)
 def is_safe_url(url):
     if not is_safe_schema(url):
         return False
+
     result = urlparse(url)
+
     return is_safe_domain(result.netloc)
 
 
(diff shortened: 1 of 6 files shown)


ocean_provider/util_url.py
```diff
@@ -0,0 +1,73 @@
...
+        return False
+    result = urlparse(url)
+    return is_safe_domain(result.netloc)
+
+
+def is_safe_schema(url):
+    try:
+        result = urlparse(url)
+        return all([result.scheme, result.netloc, result.path])
+    except:
```
calina-c, line 22: can we identify an exception class here, or a list of exception classes to be caught?

ocean_provider/util_url.py
```diff
@@ -0,0 +1,73 @@
...
+        logger.info(
+            "[i] Cannot get A record for domain '%s': %s\n" % (domain, e))
+    try:
+        ip_v6_records = DNS_RESOLVER.resolve(domain, "AAAA", search=True)
+    except Exception as e:
+        ip_v6_records = None
+        logger.info(
+            "[i] Cannot get AAAA record for domain '%s': %s\n" % (domain, e))
+
+    if verify_dns_records(domain, ip_v4_records, "A") or verify_dns_records(domain, ip_v6_records, "AAAA"):
```
calina-c, line 41: are we verifying they are valid or invalid?

ocean_provider/util_url.py
```diff
@@ -0,0 +1,73 @@
...
+        return False
+    return True
+
+
+def verify_dns_records(domain, records, type):
```
calina-c, line 46: can you rename the "type" parameter into something more explicit? I also think "type" is a reserved keyword and might get us into trouble with Python

ocean_provider/util_url.py
```diff
@@ -0,0 +1,73 @@
...
+                    type, domain, value))
+                error_detected = False
+            if ip.is_global:
+                error_detected = False
+            if error_detected == True:
+                logger.error("[!] DNS record type '%s' for domain name '%s' resolve to a non public IP address '%s'!" % (
+                    type, domain, value))
+
+        except ValueError:
+            error_detected = True
```
calina-c, line 71: If we move this (inside the "for" loop) logic to a separate function, the syntax can become more clear and concise. Instead of setting the flag, we can directly return False/(True?) when an error is encountered, and skip the rest of the execution.

calina-c (changes requested):

ocean_provider/util_url.py
alexcos20: do we need to ? after all, we need to make sure that a URL is valid (has a schema, a netloc (domain),and a path.
If it doesn't, we will reject it anyway

ocean_provider/util_url.py
alexcos20: We check them in order to make sure that the domain does not resolve to a private class of ips  (some of them are allowed if allow_non_public_ip flag is set.

ocean_provider/util_url.py
alexcos20: type is the record type.  we need to check for both "A"  (ipv4) and "AAAA" (ipv6)

ocean_provider/util_url.py
alexcos20: but what if a domain has multiple ip addresses? we need to make sure that all of them are safe.
Feel free to change

ocean_provider/util_url.py
calina-c: then how about record_type?

ocean_provider/util_url.py
calina-c: I mean the `except :` thing is not usually recommended.  The fewer of these the better?

ocean_provider/util_url.py
```diff
@@ -0,0 +1,100 @@
...
+
+    try:
+        ip = ipaddress.ip_address(value)
+        # noqa See https://docs.python.org/3/library/ipaddress.html#ipaddress.IPv4Address.is_global
+        if ip.is_private or ip.is_reserved or ip.is_loopback:
+            if allow_non_public_ip:
+                logger.warning(
+                    f"[!] DNS record type {record_type} for domain name "
+                    f"{domain} resolves to a non public IP address {value}, "
+                    "but allowed by config!"
```
alexcos20, line 86:
```suggestion
                    "but allowed by config!"
                    return True
```

calina-c (approved):

PR #77, opened 2021-02-01
Corrects docs for Swagger representation.

### LINKED ISSUE #76: API spec out of date resulting in no API docs
The spec used to render the API docs is outdated, making https://docs.oceanprotocol.com/references/provider/ useless. Additionally, there's a `./API.md` in the repo outlining API endpoints out of sync with the code comments.

There should be only 1 source of truth for all API endpoints and that should be the code with their comments, which then gets transformed into the spec mentioned above.

calina-c:
Fixes #76.

Changes proposed in this PR:

- add correct docs formatting for swagger

### DIFF
--- a/ocean_provider/routes.py
+++ b/ocean_provider/routes.py
@@ -131,16 +131,14 @@ def simple_flow_consume():
 
 @services.route('/encrypt', methods=['POST'])
 def encrypt():
-    """
-    Encrypt document using the Provider's own symmetric key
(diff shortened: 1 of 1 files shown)


kremalicious (approved):
oh, great find! The newlines were at fault

calina-c:
> oh, great find! The newlines were at fault

and the missing dividers :D

PR #82, opened 2021-02-05
Add Black autoformatting and pre-commit hooks. Separate compute and consume routes.

calina-c:
What the title says :)

### DIFF
--- a/.github/workflows/black.yml
+++ b/.github/workflows/black.yml
@@ -1,0 +1,13 @@
+name: black
+
+on: [push]
+
+jobs:
+  black:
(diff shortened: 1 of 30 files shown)


calina-c:
@alexcos20 The failures on Codacity are due to the special formatting required by Swagger to autogenerate the docs. I did not want to change those. I think we can merge without the full Codacity approval, in this case. I also chose to separate consume and compute routes in different files, per @ssallam's suggestion and to prepare the ground for upcoming compute changes.

alexcos20 (approved):
lgtm

PR #90, opened 2021-02-09
Remote asset

### LINKED ISSUE #86: Remote asset
Overview:

Alice published an algo (https://market.oceanprotocol.com/asset/did:op:A9E3385859e59407De50ed4787E788DB2b99FB58) using their own provider(Alice's provider)

Alice published an compute asset (https://market.oceanprotocol.com/asset/did:op:70859f3393b3405443Af3066C1B3b179a8b03a94) using our provider (OPF)

Bob tries to run a compute job with that algo. Provider fails, because it tries to decrypt the url for all assets

Solution:
For every compute asset (dataset, algo, etc), provider should check first if that asset is served by it or is it another provider (ocean.py has a new function for this (https://github.com/oceanprotocol/ocean.py/blob/master/ocean_lib/data_provider/data_service_provider.py#L371)

If it's served by local provider, then we should have the url  in the compute workflow.
Otherwise, instead of url we should have the remote endpoint and txid for consumption.

So, our workflow will look like:
```json
"workflow":{
        "stages": [
          {
            "index": 0,
            "input": [
              {
               "index": 0,
                "id": "did:op:87bdaabb33354d2eb014af5091c604fb4b0f67dc6cca4d18a96547bffdc27bcf",
                "url": [
                  "https://data.ok.gov/sites/default/files/unspsc%20codes_3.csv"
                ]
                
              },
              {
                "index": 1,
                "id": "did:op:1384941e6f0b46299b6e515723df3d8e8e5d1fb175554467a1cb7bc613f5c72e",
                "remote": {
		  "serviceEndpoint":"https://remoteprovider.com",
		  "txId":"0x1234",
                  "serviceIndex":1
		}
              }
            ],
}
]
}
```
With that, the CtD enviroment will be able to connect to the remote provider and consume

calina-c:
Fixes #86.

### DIFF
--- a/ocean_provider/serializers.py
+++ b/ocean_provider/serializers.py
@@ -5,7 +5,6 @@ from ocean_provider.util import (
     validate_order,
     validate_transfer_not_used_for_other_service,
 )
-from ocean_provider.util_url import get_base_url
 from ocean_provider.utils.basics import get_asset_from_metadatastore, get_config
 from ocean_utils.agreements.service_agreement import ServiceAgreement
 from ocean_utils.agreements.service_types import ServiceTypes
@@ -75,9 +74,7 @@ class StageAlgoSerializer:
         dict_template["id"] = algorithm_did
         dict_template["rawcode"] = ""
 
-        if get_base_url(service.service_endpoint) == get_base_url(
-            get_config().provider_url
-        ):
+        if get_config().provider_address == self.provider_wallet.address:
             dict_template["url"] = get_asset_url_at_index(
                 0, algo_asset, self.provider_wallet
             )
--- a/ocean_provider/serializers.py
+++ b/ocean_provider/serializers.py
@@ -1,4 +1,3 @@
-import requests
 from eth_utils import add_0x_prefix
 from ocean_provider.util import (
     get_asset_url_at_index,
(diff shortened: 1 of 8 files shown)


ocean_provider/serializers.py
```diff
@@ -0,0 +1,94 @@
...
+            return dict(
+                {
+                    "id": "",
+                    "url": algorithm_meta.get("url"),
+                    "rawcode": algorithm_meta.get("rawcode"),
+                    "container": algorithm_meta.get("container"),
+                }
+            )
+
+        msg = "algorithm_did requires both algorithm_token_address and algorithm_tx_id."
```
ssallam, line 44: it's better to use the actual names in the message: `algorithmDataToken` and `algorithmTransferTxId`

calina-c: done!

ocean_provider/util_url.py
```diff
@@ -34,6 +35,21 @@ def is_ip(address):
...
 
+def is_this_same_provider(url):
+    result = urlparse(url)
+
+    try:
+        return (
+            DataServiceProvider().get_provider_address(
+                f"{result.scheme}://{result.netloc}/"
+            )
+            == get_provider_wallet().address
```
ssallam, line 46: would it be safer to lowercase both sides?

ssallam (approved):
LGTM.

PR #92, opened 2021-02-09
prevent assigning None to fileInfo.contentLength

### LINKED ISSUE #91: fileinfo: contentLength key is always present in response
When checking a URL where `content-length` can't be detected, the response still adds `contentLength: null` which can break Aquarius metadata validation later on in the user flow.

screen_shot_2021-02-09_at_18 28 49

When no `content-length` can be detected, it should not be added at all to the response. Test with `https://oceanprotocol.com` as the file url.

This addresses the root cause of https://github.com/oceanprotocol/market/issues/367

ssallam:
Fixes #91

Changes proposed in this PR:

- fix error when checking for unsafe url
- prevent assigning None to contentType or contentLength


calina-c (dismissed):

calina-c:
I just noticed the `#` in the branch name, which breaks Docker builds. Do not merge this until then :D

PR #94, opened 2021-02-10
Additional input

### LINKED ISSUE #50: Provider:  multiple files for compute dataset & algo
- Add additionalInput parameter, which contains an additional list of assets, that will be added to workflow input list.

  For each asset, we can have either:
     - serviceType = access  (all assets are allowed) (Check if local provider vs remote provider and fill the workflow fields accordingly 
     - serviceType = compute (the dataset needs to be served by local provider, else call will fail)
     
Structure:

 ```json
[ 
       {"did":"0x1234",  "txId":"0x876", "serviceId":1} ,
       {"did":"0x1267",  "txId":"0x345", "serviceId":1}
]
```

### PR DESCRIPTION
**Pull request by calina-c**
Fixes #50.

### CODE REVIEW

File: ocean_provider/utils/encryption.py [hunk 1]
```diff
@@ -13,10 +13,12 @@ def do_encrypt(document, wallet: Wallet = None, public_key=None):
...
     key = get_private_key(provider_wallet)
-
-    return ecies.decrypt(key.to_hex(), Web3.toBytes(hexstr=encrypted_document)).decode(
-        encoding="utf-8"
-    )
+    try:
+        return ecies.decrypt(
+            key.to_hex(), Web3.toBytes(hexstr=encrypted_document)
+        ).decode(encoding="utf-8")
+    except:  # noqa
```
ssallam, line 20: in general, should never use a bare `except`. rather have `except Exception`

ocean_provider/validation/algo.py
```diff
@@ -157,3 +197,134 @@ def validate_formatted_algorithm_dict(algorithm_dict, algorithm_did):
...
+        self.data = data
+        self.index = index
+
+    def validate(self):
+        required_keys = ["documentId", "transferTxId", "serviceId"]
+
+        for req_item in required_keys:
+            if not self.data.get(req_item) and not (
+                req_item == "serviceId" and self.data.get(req_item) == 0
+            ):
```
ssallam, line 216: if serviceId is special, it makes it simpler to handle it separately.

ocean_provider/validation/algo.py
```diff
@@ -157,3 +197,134 @@ def validate_formatted_algorithm_dict(algorithm_dict, algorithm_did):
...
+            self.asset, self.provider_wallet, config_file=app.config["CONFIG_FILE"]
+        )
+
+        if self.service.type == ServiceTypes.CLOUD_COMPUTE and not asset_urls:
+            self.error = "Services in input with compute type must be in the same provider you are calling."
+            return False
+
+        if self.service.type != ServiceTypes.CLOUD_COMPUTE and self.index == 0:
+            self.error = "Service for main asset must be compute."
+            return False
```
ssallam, line 250: since this check does not rely on asset_urls, it should occur before getting asset_urls.

ssallam (approved):
LGTM.

PR #122, opened 2021-03-31
Guess algo service in compute to data.

calina-c:


ocean_provider/validation/algo.py
```diff
@@ -132,31 +132,52 @@ def _build_and_validate_algo(self, algo_data):
                 return False
 
             try:
-                service = ServiceAgreement.from_ddo(ServiceTypes.ASSET_ACCESS, algo)
+                dt = DataToken(self.consumer_address)
+                tx_receipt = dt.get_tx_receipt(algorithm_tx_id)
```
alexcos20, line 136: what if that algorithm_tx_id does not exists? I think we should catch this

alexcos20 (approved):

ocean_provider/validation/algo.py
calina-c: If the algo does not exist, that is treated earlier. (line 124 and further). The only option is if there's no tx receipt for the algorithm, which will generate an exception. I added a more explicit message in the following commit. Let me know if more is needed.

PR #150, opened 2021-05-25
Multiple files downloaded

### LINKED ISSUE #54: Provider: ocean.py marketplace quickstart hangs when downloading >1 files
When I was working to solve "ocean.py marketplace quickstart fails" [ocean.py #89](https://github.com/oceanprotocol/ocean.py/issues/89), I encountered an issue: it would hang when trying to download >1 files.

My priority was to get the flow working. So I switched from needing 3 services to 1, and from using AWS to using github and a file that I knew to be small. That resolved my immediate priority.

While I still had it set up for 3 services, I traced the code. It seemed to be hanging inside the requests.py library, when doing chunked iterations towards downloading. It hung for >10 min. Was this a large file issue? No - Samer and Alex tell me that the file was small. 

Alex also mentioned that he has success using >1 service in compute-to-data settings.

Today in standup someone else mentioned this being a possible issue in provider.py. 

To reproduce this bug on a codebase as close-as-possible to the current codebase:
1. Branch from current ocean.py 
2. In marketplace quickstart, copy in the metadata/services info from code prior to the commit that solves ocean.py `#89` . [This commit has the relevant changes](https://github.com/oceanprotocol/ocean.py/commit/3c5e07f02ecca817c50231a93fdea2bc39fdd75d) which allowed to run to the end of step 7/7.
3. Run marketplace quickstart. There may be two outcomes: (a) it fails like I had above. (b) It doesn't fail due to other bug fixes since this issue was reported. 

TODOs for this issue:
- [x] try to reproduce, as given above. It may or may not fail, see step 3 above.
- [x] write a unit or system test on >1 file. It may or may not fail, as mentioned
- [ ] if there was an issue: fix the issue 
- [ ] as part of PR, re-deploy to pip following the deploy process [ocean.py#188](https://github.com/oceanprotocol/ocean.py/issues/118), [provider#78](https://github.com/oceanprotocol/provider/issues/78)

mariacarmina:
Fixes #54 .

Changes proposed in this PR:

- created an unit test to see if there is an issue or not;
- added a new ddo sample with multiple files.

### DIFF
--- a/tests/test_download.py
+++ b/tests/test_download.py
@@ -137,30 +137,14 @@ def test_download_multiple_files(client, publisher_wallet, consumer_wallet):
     response = client.get(download_endpoint, query_string=payload)
     assert response.status_code == 200, f"{response.data}"
 
-    payload = {
-        "documentId": ddo.did,
-        "serviceId": sa.index,
-        "serviceType": sa.type,
-        "dataToken": ddo.data_token_address,
-        "consumerAddress": consumer_wallet.address,
-        "signature": generate_auth_token(consumer_wallet),
-        "transferTxId": tx_id,
-        "fileIndex": 1,
-    }
+    payload["signature"] = generate_auth_token(consumer_wallet)
+    payload["fileIndex"] = 1
     download_endpoint = BaseURLs.ASSETS_URL + "/download"
     response = client.get(download_endpoint, query_string=payload)
     assert response.status_code == 200, f"{response.data}"
 
-    payload = {
-        "documentId": ddo.did,
-        "serviceId": sa.index,
-        "serviceType": sa.type,
-        "dataToken": ddo.data_token_address,
-        "consumerAddress": consumer_wallet.address,
-        "signature": generate_auth_token(consumer_wallet),
-        "transferTxId": tx_id,
-        "fileIndex": 2,
-    }
+    payload["signature"] = generate_auth_token(consumer_wallet)
+    payload["fileIndex"] = 2
     download_endpoint = BaseURLs.ASSETS_URL + "/download"
     response = client.get(download_endpoint, query_string=payload)
     assert response.status_code == 200, f"{response.data}"
--- a/tests/test_download.py
+++ b/tests/test_download.py
@@ -17,7 +17,7 @@ from tests.test_helpers import (
     get_dataset_with_ipfs_url_ddo,
     get_nonce,
     mint_tokens_and_wait,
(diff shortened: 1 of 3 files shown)


mariacarmina:
The last commit is the final version

tests/test_download.py
```diff
@@ -109,3 +112,58 @@ def test_initialize_on_asset_with_custom_credentials(
...
+
+    mint_tokens_and_wait(dt_token, consumer_wallet, publisher_wallet)
+
+    sa = ddo.get_service(ServiceTypes.ASSET_ACCESS)
+    tx_id = send_order(client, ddo, dt_token, sa, consumer_wallet)
+
+    # Consume using url index and auth token
+    # (let the provider do the decryption)
+    payload = {
+        "documentId": ddo.did,
```
calina-c, line 129: Since these payloads only differ by fileIndex, can we please not copy/paste and assign? Instead, let's make one request -> change the body -> do the second etc.

tests/test_helpers.py
```diff
@@ -176,6 +176,16 @@ def get_dataset_ddo_with_access_service(client, wallet):
     return get_registered_ddo(client, wallet, metadata, service_descriptor)
 
 
+def get_dataset_ddo_with_multiple_files(client, wallet):
+    metadata = get_sample_ddo_with_multiple_files()["service"][0]["attributes"]
+    metadata["main"]["files"][0]["checksum"] = str(uuid.uuid4())
```
calina-c, line 181: Can we do a range(3) here instead of 3 separate lines?
I.e. for i in range(3): metadata["main"]["files"][i] ...

calina-c (changes requested):
Just a few minor tweaks suggested. The functionality is ok.

tests/test_download.py
```diff
@@ -18,6 +18,9 @@
     get_nonce,
     mint_tokens_and_wait,
     send_order,
+    get_algorithm_ddo,
```
calina-c, line 21: are all these 3 imports needed?

tests/test_download.py
calina-c: @mariacarmina I mean you can change just the keys that need to be changed. I think fileIndex and signature. The rest of the payload stays the same, so no need to redeclare it. Just update the 2 changed keys.

calina-c (changes requested):

calina-c (approved):

PR #153, opened 2021-06-09
get minter from contract

### LINKED ISSUE #154: DT Contract should be the source of truth, not ddo when taking dt minter
**Describe the bug**
When getting the minter of a DT, we use the DDO information and check if isMinter.

**Expected behavior**
We should take the minter from the DT contract

alexcos20:
Closes #154

### DIFF
--- a/ocean_provider/utils/basics.py
+++ b/ocean_provider/utils/basics.py
@@ -27,7 +27,7 @@ def get_provider_wallet():
 
 def get_datatoken_minter(asset, datatoken_address):
     dt = DataToken(datatoken_address)
-    publisher = dt.contract_concise.minter()
+    publisher = dt.minter()
     return publisher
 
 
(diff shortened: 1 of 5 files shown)


ocean_provider/utils/basics.py
```diff
@@ -26,15 +26,8 @@ def get_provider_wallet():
...
-    publisher = Web3Provider.get_web3().toChecksumAddress(asset.publisher)
     dt = DataToken(datatoken_address)
-
-    assertion_message = (
-        f"ddo publisher {publisher} is not the current "
-        f"minter for the DataToken contract at {datatoken_address}."
-    )
-    assert dt.contract_concise.isMinter(publisher), assertion_message
-
+    publisher = dt.contract_concise.minter()
```
trentmc, line 30: Nice catch. Even simpler:

```python
publisher = dt.minter()
```

trentmc (changes requested):

trentmc (approved):

calina-c (approved):

PR #155, opened 2021-06-09
Integrate provider with RBAC server

### LINKED ISSUE #152: Integrate Provider with RBAC server

mariacarmina:
Fixes #152 .

Changes proposed in this PR:

- integrate the RBAC server logic in provider.


alexcos20:
RBAC is another condition.  and an OR.

So if you have RBAC_URL, you are going to follow old validations flow + check RBAC.
If there is no RBAC -> just follow old validation flow

calina-c:
> RBAC is another condition. and an OR.
> 
> So if you have RBAC_URL, you are going to follow old validations flow + check RBAC.
> If there is no RBAC -> just follow old validation flow

@mariacarmina no problem here :) Simply instead of declaring self.validator, append to a list of validators. And overwrite the part where self.validator.fails() is called to iterate over all the existing validators. Minimal changes needed.

mariacarmina:
> > RBAC is another condition. and an OR.
> > So if you have RBAC_URL, you are going to follow old validations flow + check RBAC.
> > If there is no RBAC -> just follow old validation flow
> 
> @mariacarmina no problem here :) Simply instead of declaring self.validator, append to a list of validators. And overwrite the part where self.validator.fails() is called to iterate over all the existing validators and check that none of them fails. Minimal changes needed

OK, I will refactor it. Thanks @calina-c

ocean_provider/validation/requests.py
```diff
@@ -48,6 +68,103 @@ def __init__(
...
+        return self._assets
+
+    @property
+    def request(self):
+        return self._request
+
+    def passes(self):
+        return True
+
+    def fails(self):
```
calina-c, line 114: this needs to be adapted. Please remove the passes function, since it is not needed at all. And implement the contents for the fails function. This is now only a placeholder.

ocean_provider/validation/requests.py
```diff
@@ -48,6 +68,103 @@ def __init__(
...
+            "component": self.component,
+            "providerAddress": get_provider_wallet().address,
+            "credentials": self.credentials,
+        }
+        payload.update(getattr(self, f"build_{self.action}_payload")())
+        return payload
+
+    def build_encryptUrl_payload(self):
+        message = "encryptUrl" + json.dumps(self.credentials)
+        signature = sign_hash(
```
calina-c, line 129: I think there's a signing function somewhere. If not, you can create it. Can we also add the provider wallet to the list of properties, to avoid using the function every time?

ocean_provider/validation/requests.py
```diff
@@ -48,6 +78,150 @@ def __init__(
...
+            "ComputeRequest": "compute",
+            "ComputeStartRequest": "compute",
+        }
+        self._action = action_mapping[request_name]
+        self._credentials = {
+            "type": "address",
+            "address": get_provider_wallet().address,
+        }
+        self._component = "provider"
+        self._assets = assets
```
calina-c, line 105: I think you can simplify these to one-liners using the python ternary operator: self._assets = assets if assets else []

ocean_provider/validation/requests.py
```diff
@@ -48,6 +78,150 @@ def __init__(
...
+        self._algorithms = algorithms
+        if self.assets is None:
+            self._assets = []
+        if self.algorithms is None:
+            self._algorithms = []
+        self._compute_dids = {
```
calina-c, line 111: do we need these in multiple places? If not, the logic can belong to a child class.

ocean_provider/validation/requests.py
```diff
@@ -48,6 +78,150 @@ def __init__(
...
+            "algos": [
+                {
+                    "did": algorithm.did,
+                    "serviceId": ServiceTypesIndices.DEFAULT_COMPUTING_INDEX,
+                }
+                for algorithm in self.algorithms
+            ],
+        }
+
+    @property
```
calina-c, line 128: I think these properties overcrowd the class and there is not much added value. I think you can take the following action:
- look into which of these are "public" (used in other places) and remove the underscore. I'm not a huge fan of the underscore either way.
- maybe keep provider_address as a property? But even that one, I'd rather have it as self.provider_address inside init

ocean_provider/validation/requests.py
```diff
@@ -48,6 +78,150 @@ def __init__(
...
+    def algorithms(self):
+        return self._algorithms
+
+    def fails(self):
+        return False
+
+    def msg_hash(self, message: str):
+        return hashlib.sha256(message.encode("utf-8")).hexdigest()
```
calina-c, line 164: I am 100% sure this logic exists somewhere in utils, no need to implement it as a function of RBAC classes, since it is not its concern.

ocean_provider/validation/requests.py
```diff
@@ -48,6 +78,150 @@ def __init__(
...
+
+    def build_payload(self):
+        payload = {
+            "eventType": self.action,
+            "component": self.component,
+            "providerAddress": self.provider_address,
+            "credentials": self.credentials,
+        }
+        payload.update(getattr(self, f"build_{self.action}_payload")())
```
calina-c, line 173: can you please leave a comment here e.g. "builds actions like build_encrtyptUrl_payload to update the dictionary with request-specific key-values." (or something similar)

ocean_provider/validation/requests.py
```diff
@@ -48,6 +78,150 @@ def __init__(
...
+        signature = sign_hash(self.msg_hash(message), get_provider_wallet())
+
+        return {"signature": signature}
+
+    def build_initialize_payload(self):
+        message = "initialize" + json.dumps(self.credentials)
+        signature = sign_hash(self.msg_hash(message), get_provider_wallet())
+        return {
+            "signature": signature,
+            "dids": [
```
calina-c, line 187: either initialize it here, or keep it in the init, but not both places. See my comment on scoping (if this is something common to more requests, init will do. Otherwise, keep it only here)

ocean_provider/validation/requests.py
```diff
@@ -48,6 +78,150 @@ def __init__(
...
+                for asset in self.assets
+            ],
+        }
+
+    def build_access_payload(self):
+        message = "access" + json.dumps(self.credentials)
+        signature = sign_hash(self.msg_hash(message), get_provider_wallet())
+        return {
+            "signature": signature,
+            "dids": [
```
calina-c, line 201: same here.

tests/test_RBAC.py
```diff
@@ -0,0 +1,138 @@
...
+from tests.test_helpers import (
+    mint_tokens_and_wait,
+    get_dataset_ddo_with_access_service,
+    send_order,
+)
+
+
+def test_null_validator():
+    with pytest.raises(RequestNotFound):
+        RBACValidator()
```
calina-c, line 24: ah, please implement RBACValidator as an Abstract Base Class just to be sure. Keep the tests, but alter it to take a gibberish request instead.

tests/test_RBAC.py
```diff
@@ -0,0 +1,138 @@
...
+            "documentId": ddo.did,
+            "serviceId": sa.index,
+            "serviceType": sa.type,
+            "dataToken": ddo.data_token_address,
+            "consumerAddress": consumer_wallet.address,
+        }
+    )
+
+    request_url = (
+        init_endpoint + "?" + "&".join([f"{k}={v}" for k, v in payload.items()])
```
calina-c, line 71: Please don't do this. Did some of this remain in the code base? I remember refactoring all the requests as proper GET requests, not hackish stuff with URL formation. Did I miss some places, where did you get this from? Please fix there too.

tests/test_RBAC.py
```diff
@@ -0,0 +1,138 @@
...
+    assert payload["credentials"] == validator.credentials
+
+
+def test_initialize_request_payload(client, publisher_wallet, consumer_wallet):
+    ddo = get_dataset_ddo_with_access_service(client, publisher_wallet)
+    dt_contract = DataToken(ddo.data_token_address)
+    sa = ddo.get_service(ServiceTypes.ASSET_ACCESS)
+    mint_tokens_and_wait(dt_contract, consumer_wallet, publisher_wallet)
+
+    payload = dict(
```
calina-c, line 60: no need for the dict conversion, I think. This is already a dict by definition.

tests/test_RBAC.py
```diff
@@ -0,0 +1,138 @@
...
+            "dataToken": ddo.data_token_address,
+            "consumerAddress": consumer_wallet.address,
+        }
+    )
+
+    request_url = (
+        init_endpoint + "?" + "&".join([f"{k}={v}" for k, v in payload.items()])
+    )
+    document = {
+        "url": request_url,
```
calina-c, line 74: actually something is not right. Please handle the rest of the code review and let's discuss how request_url ended up inside `document`... ?!

tests/test_download.py
```diff
@@ -71,6 +71,17 @@ def test_empty_payload(client):
     assert consume.status_code == 400
 
 
+def test_initialize_on_dataset_with_access_service(
```
calina-c, line 74: is this still needed? I remember it was a temporary test.

ocean_provider/validation/requests.py
mariacarmina: I didn't see it. I put in utils btw.

mariacarmina:
For the last commit, I created a prototype of a possible idea to actually give the payload as a parameter to the validator class, because I needed the `additionalInputs` key. And maybe this idea could remove the `document` dict from each `test_build_payload`.

ocean_provider/validation/RBAC.py
```diff
@@ -0,0 +1,170 @@
+#
+# Copyright 2021 Ocean Protocol Foundation
+# SPDX-License-Identifier: Apache-2.0
+#
+
+import json
```
calina-c, line 6: can you please order these imports?

ocean_provider/utils/util.py
```diff
@@ -37,6 +38,10 @@ def get_request_data(request):
     return request.args if request.args else request.json
 
 
+def msg_hash(message: str):
```
calina-c, line 41: am I confusing this with aquarius? I am sure such a function already exists. It has to... or there are cases where the same logic is used, where we need to use this function instead.

either way, please add a test for this function specifically in test_utils.

ocean_provider/validation/RBAC.py
```diff
@@ -0,0 +1,170 @@
...
+        self._credentials = {
+            "type": "address",
+            "address": get_provider_wallet().address,
+        }
+        self._component = "provider"
+        self._provider_address = get_provider_wallet().address
+        self._assets = assets if assets else []
+        self._algorithms = algorithms if algorithms else []
+
+    @property
```
calina-c, line 46: I left a comment previously to discuss these property things. That was not addressed, but marked resolved. Please handle my original comment.

ocean_provider/validation/RBAC.py
```diff
@@ -0,0 +1,170 @@
...
+    def additional_inputs(self):
+        return self._payload["additionalInputs"]
+
+    def fails(self):
+        return False
+
+    def get_compute_dict(self):
+        return {
+            "dids": [
```
calina-c, line 83: you can start with the rest of the dictionary, then simply add compute_dict["dids"] = self.get_dids() to avoid repetition

ocean_provider/validation/RBAC.py
```diff
@@ -0,0 +1,170 @@
...
+        signature = sign_hash(msg_hash(message), get_provider_wallet())
+        return {
+            "signature": signature,
+            "dids": dids,
+        }
+
+    def build_compute_payload(self):
+        dids = self.get_compute_dict()["dids"]
+        algos = self.get_compute_dict()["algos"]
+        additional_dids = self.get_compute_dict()["additionalDids"]
```
calina-c, line 156: why is get_compute_dict a different function?
- it is only used here
- and it is used piecemeal...

move the functionality here, remove the function.
dids will be self.get_dids either way
you just need to add algos and additional dids.

ocean_provider/validation/RBAC.py
```diff
@@ -0,0 +1,170 @@
...
+        request=None,
+        payload: dict = None,
+        assets: Optional[list] = None,
+        algorithms: Optional[list] = None,
+    ):
+        self._request = request
+        self._payload = payload if payload else dict()
+        if not request:
+            raise RequestNotFound("Request not found.")
+        action_mapping = {
```
calina-c, line 29: please define the action mapping as a static property on the class. Use the full dictionary when mappping the action, and use the `keys()` when adding RBACValidator to the list of validators (right now it is an identical usage, but only with the requst names, rewritten in plain)

ocean_provider/validation/requests.py
```diff
@@ -21,15 +28,36 @@ class CustomJsonRequest(JsonRequest):
...
-            rules=self.rules(),
-            messages={
-                "signature.signature": "Invalid signature provided.",
-                "signature.download_signature": "Invalid signature provided.",
-            },
-            request=request,
+        class_name = self.__class__.__name__
+        self._validators = list()
+        if os.getenv("RBAC_SERVER_URL") and class_name in [
+            "InitializeRequest",
```
calina-c, line 34: i.e. here we can reuse the action mapper keys instead of rewriting this array by hand

calina-c (changes requested):

ocean_provider/validation/RBAC.py
```diff
@@ -0,0 +1,170 @@
...
+
+    @property
+    def algorithms(self):
+        return self._algorithms
+
+    @property
+    def additional_inputs(self):
+        return self._payload["additionalInputs"]
+
+    def fails(self):
```
calina-c, line 78: please remember this TODO. Have you discussed the RBAC response structure with @jamiehewitt15?

btw. please take into account that, if the answer from the RBAC server is variable (e.g. they give us an error message), we need to configure that in the RBAC Validator's `message()` function (probably set it as a property when doing the request and having the function return the property). Otherwise we can hardcode the `messages` function to simply return "RBAC authorization failed." or something similar.

ocean_provider/validation/RBAC.py
mariacarmina: They are different dicts. The `serviceId` is different in this case, that's why I created 2 functions and `get_compute_dict` is just for the compute service.

ocean_provider/validation/RBAC.py
jamiehewitt15: The error message you get back will be in the normal [HTTP response format](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status)

In the market we just log the error message to the console: https://github.com/oceanprotocol/market/blob/04d505da42f84839e752b4ffa5d4f3943f563faf/src/utils/rbac.ts#L30

jamiehewitt15: The body of the response will only be `true` or `false`

calina-c: yes, but the response itself, is it e.g.
```{"result": "true", "message": "Something..."}```
OR
```true```
OR
something else?

Comment (jamiehewitt15) [reply ocean_provider/validation/RBAC.py:78]: The second one, just `true` or `false`

Comment (jamiehewitt15) [reply ocean_provider/validation/RBAC.py:78]: By the way, I'm not sure if you're doing this already but for testing you can start the RBAC server on barge with `./start_ocean.sh --with-rbac`

There are some preconfigured accounts with roles that work barge: https://github.com/oceanprotocol/barge#rbac-server

File: ocean_provider/utils/util.py (hunk 16 above)
Comment (mariacarmina) [reply ocean_provider/utils/util.py:41]: I didn't see anything similar.

File: ocean_provider/validation/RBAC.py (hunk 17 above)
Comment (mariacarmina) [reply ocean_provider/validation/RBAC.py:46]: I didn't marked that as resolved, is outdated due to the new pushed commits.

File: ocean_provider/validation/RBAC.py (hunk 22 above)
Comment (calina-c) [reply ocean_provider/validation/RBAC.py:55]: That sounds perfect!
CC: @mariacarmina Add the barge option in tests and it should work properly :)

File: ocean_provider/validation/RBAC.py (hunk 19 above)
Comment (calina-c) [reply ocean_provider/validation/RBAC.py:133]: I understand the dids part is different, ok 👍 .
Still, please move the functionality from compute_dict here and do not build it as a dict if you don't use it as a dict. If the only difference between the return dict and the compute dict that you build is one key, why do you not add it with name_of_dict["extra_key"] = variable ?

File: tests/test_RBAC.py [hunk 23]
```diff
@@ -0,0 +1,173 @@
...
+from tests.helpers.compute_helpers import build_and_send_ddo_with_compute_service
+from tests.test_helpers import (
+    get_dataset_ddo_with_access_service,
+    mint_tokens_and_wait,
+    send_order,
+)
+
+
+def test_null_validator():
+    with pytest.raises(RequestNotFound):
```
calina-c, line 25: why does this still raise "RequestNotFound"? RBAC Validator is an abc now, right? It shouldn't be instantiated

tests/test_RBAC.py
```diff
@@ -0,0 +1,173 @@
...
+            "contentType": "text/csv",
+            "encoding": "UTF-8",
+            "compression": "zip",
+        }
+    ]
+    req = {"document": json.dumps(document[0])}
+    validator = RBACValidator(request_name="EncryptRequest", request=req)
+    payload = validator.build_payload()
+    assert validator.request == req
+    assert payload
```
calina-c, line 48: no need to assert the payload if you are asserting its content. Instead of checking against the validator.action (component, credentials...), can we test the values themselves?

tests/test_RBAC.py
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    req = {"document": json.dumps(payload)}
+    validator = RBACValidator(
+        request_name="InitializeRequest", request=req, payload=payload, assets=[ddo]
+    )
+    validator_payload = validator.build_payload()
+    assert validator.request == req
+    assert validator_payload
+    assert validator_payload["eventType"] == validator.action
```
calina-c, line 75: same here and the folowing tests...

ocean_provider/validation/RBAC.py
```diff
@@ -0,0 +1,146 @@
...
+        self.action = action_mapping[request_name]
+        self.credentials = {
+            "type": "address",
+            "address": get_provider_wallet().address,
+        }
+        self.component = "provider"
+        self.provider_address = get_provider_wallet().address
+        self.assets = assets if assets else []
+        self.algorithms = algorithms if algorithms else []
+        if request_name in ["ComputeRequest", "ComputeStartRequest"]:
```
calina-c, line 40: it seems we don't have an understanding of what belongs in the parent class, and what should be defined in each functiono specifically. if the request is one of those, please compute the additional_inputs variable inside the function, using self.payload which is generic to all requests.

ocean_provider/validation/RBAC.py
```diff
@@ -0,0 +1,146 @@
...
+        self,
+        request_name=None,
+        request=None,
+        payload: Optional[dict] = None,
+        assets: Optional[list] = None,
+        algorithms: Optional[list] = None,
+    ):
+        self.request = request
+        self.payload = payload if payload else dict()
+        if not request:
```
calina-c, line 28: is this in case of no request, or no "request_name"? I am a bit confused about the logic.

ocean_provider/validation/RBAC.py
```diff
@@ -0,0 +1,146 @@
...
+    @staticmethod
+    def get_action_mapping():
+        action_mapping = {
+            "EncryptRequest": "encryptUrl",
+            "InitializeRequest": "initialize",
+            "DownloadRequest": "access",
+            "ComputeRequest": "compute",
+            "ComputeStartRequest": "compute",
+        }
+        return action_mapping
```
calina-c, line 52: you can return the dict directly, without assigning it.

ocean_provider/validation/RBAC.py
calina-c: this is still not handled.

ocean_provider/validation/RBAC.py
```diff
@@ -0,0 +1,146 @@
...
+                {
+                    "did": additional_input["documentId"],
+                    "serviceId": additional_input["serviceId"],
+                }
+                for additional_input in self.additional_inputs
+            ],
+        }
+
+    def get_dids(self):
+        return {
```
calina-c, line 83: I don't understand why these are dictionaries with a single key? And used by accessing that key?
I mean you have `get_dids(self): return {"dids": { "foo": "bar"} }` and then you use it as `get_dids()["dids"]`
Why not return {"foo": "bar"}  and use it directly?

The same goes for the compute dict, which again I did not understand why is used in the exact same way.

Also, why not paremeterise the serviceId value? You could reuse like this: get_dids(ServiceTypesIndices.DEFAULT_ACCESS_INDEX) or with the other index. i.e. a parameter called serviceTypeIndex

calina-c (changes requested):

ocean_provider/validation/RBAC.py
calina-c: remove None opption from request and request_name.
Throw RequestNotFound if request name not in action_mapping.

+ Please add a test for this

tests/test_RBAC.py
calina-c: please transform this into the proper RequestNotFound test (missing action/request_name from action mapping)

ocean_provider/validation/RBAC.py
```diff
@@ -0,0 +1,145 @@
...
+            "InitializeRequest": "initialize",
+            "DownloadRequest": "access",
+            "ComputeRequest": "compute",
+            "ComputeStartRequest": "compute",
+        }
+
+    def fails(self):
+        return False
+
+    def get_request_dict(self):
```
calina-c, line 47: What is this, exactly?

ocean_provider/validation/RBAC.py
```diff
@@ -0,0 +1,145 @@
...
+        return {
+            "signature": signature,
+            "dids": dids,
+        }
+
+    def build_compute_payload(self):
+        dids = self.get_dids(ServiceTypesIndices.DEFAULT_COMPUTING_INDEX)
+        algos = self.get_algos()
+        additional_dids = (
+            self.get_additional_dids() if self.get_additional_dids() else []
```
calina-c, line 121: Please add the if inside the function itself, to keep things cleaner. E.g. the first statement inside the get_additional_dids function can be precisely to return [] if nothing exists

ocean_provider/validation/RBAC.py
```diff
@@ -0,0 +1,145 @@
...
+        )
+        if not additional_dids:
```
calina-c, line 123: The only thing that differs between the branches of these if, is whether additional dids exist or not. I am sure you can write this in a more elegant way.

tests/test_RBAC.py
```diff
@@ -0,0 +1,216 @@
...
+    }
+    req = {"document": json.dumps(document)}
+    validator = RBACValidator(request_name="EncryptRequest", request=req)
+    payload = validator.build_payload()
+    assert validator.request == req
+    assert payload["eventType"] == "encryptUrl"
+    assert payload["component"] == "provider"
+    assert payload["credentials"] == {
+        "type": "address",
+        "address": validator.provider_address,
```
calina-c, line 52: Careful here. You are validating that the provider_address from the validator is in fact the provider_address from the validator. Use the fixture (there has to be a provider address fixture somewhere) to check this

tests/test_RBAC.py
```diff
@@ -0,0 +1,216 @@
...
+    validator = RBACValidator(request_name="InitializeRequest", request=req)
+    payload = validator.build_payload()
+    assert validator.request == req
+    assert payload["eventType"] == "initialize"
+    assert payload["component"] == "provider"
+    assert payload["credentials"] == {
+        "type": "address",
+        "address": validator.provider_address,
+    }
+    assert payload["dids"][0]["did"] == validator.get_dids(sa.index)[0]["did"]
```
calina-c, line 80: You are again checking that a = a. Use the did that you just defined a few lines above this.

calina-c: Please extend this comment to the usage of serviceId etc. And change accordingly for all the tests file.

calina-c (changes requested):

ocean_provider/validation/RBAC.py
mariacarmina: The request comes in this format: `{document: '<the actual request>'}`, so the value from the `document` key needs to be a dict instead of a string (I parse it with `json.dumps` in the tests), so basically that function converts from string to json.

calina-c: This doesn't seem to do what its name says. Is it also used only once? I am confused, let's have a call later and discuss.

ocean_provider/validation/RBAC.py
mariacarmina: And the signature will differ, too.

calina-c:
Note: Up until this point, the main functionality of building RBAC payloads is done. Please consider that part approved, but we won't merg it until we have the next step -> the remaining task is to implement the `fails` function and make the actual call to RBAC.

Difficulty: testing with the RBAC server, can we make an integration test? Otherwise, let's mock requests.get/ requests.post (see what the actual RBAC needs).

mariacarmina:
I renamed `requests.py`, the file inside `validation` directory, `provider_requests.py` to avoid the conflicts with the actual `requests` library from py. (And somehow that triggered codeclimate, too.)

mariacarmina:
I attach here the stats after a full tox running 
image

ocean_provider/validation/RBAC.py
```diff
@@ -42,7 +44,9 @@ def get_action_mapping():
         }
 
     def fails(self):
-        return False
+        payload = self.build_payload()
+        response = requests.post(f"http://{os.getenv('RBAC_SERVER_URL')}", json=payload)
```
calina-c, line 48: - the RBAC_SERVER_URL must come already prefixed! do not format it here (do not prepend http://)
- bool(response) will not do what you think it does. It will return True for all responses with a 200 status code, even if their content is "false". Use json.loads(). Have a look in the equivalent PR I made for aquarius.

requirements_dev.txt
```diff
@@ -9,3 +9,16 @@
 # and also the 'dev' list in the extras_require dict
 
 -e .[dev]
+
+pytest~=6.2.4
```
calina-c, line 13: please do not add these here. If there is something you need to add, add it to setup.py. This looks like something pyCharm suggested. But we don't code around an IDE, we code around Python. Please remove these. If anything needs to be added for the sake of functionality, please do so in setup.py. But don't do it just to oblige your IDE.

calina-c (changes requested):
A few more changes and I promise it's over :) Please handle the tests as well.

ocean_provider/validation/provider_requests.py
```diff
@@ -21,15 +26,30 @@ class CustomJsonRequest(JsonRequest):
...
+                    "signature.download_signature": "Invalid signature provided.",
+                },
+                request=request,
+            )
         )
 
+    def validate(self):
+        for validator in self._validators:
+            if validator.fails():
+                raise ValidationException(validator.messages())
```
calina-c, line 50: please implement the messages function in the validator :)
a simple statement to return "RBAC Validation failed" would be enough.

calina-c (approved):

calina-c:
🥇

calina-c (approved):

PR #172, opened 2021-06-30
Remove provider keyfile.

### LINKED ISSUE #151: README workflow fails
# Error

# Analysis

The issue is that the README workflow uses a .env file that defines the `PROVIDER_KEYFILE` envvar but, inside the code, `basics.py::get_provider_wallet()` only reads the `PROVIDER_PRIVATE_KEY` envvar.
I believe get_provider_wallet() should check both `PROVIDER_KEYFILE` and `PROVIDER_PRIVATE_KEY` and that the team should decide which one takes precedence.

calina-c:
Closes #151.


trentmc:
You'll need to update `provider/.env.example` too, it holds `PROVIDER_KEYFILE` and more.

calina-c:
> You'll need to update `provider/.env.example` too, it holds `PROVIDER_KEYFILE` and more.

Done, thanks for noticing!

DMats (approved):

PR #175, opened 2021-07-05
Upgrades ocean.py to 0.5.26.

### LINKED ISSUE #174: Remove Web3Provider
`Web3Provider` will be removed from ocean.py in Issue https://github.com/oceanprotocol/ocean.py/issues/330 (PR https://github.com/oceanprotocol/ocean.py/pull/363) and released shortly thereafter.

In preparation, usage of `Web3Provider` should be removed from provider.

calina-c:
Closes #174

### DIFF
--- a/ocean_provider/utils/basics.py
+++ b/ocean_provider/utils/basics.py
@@ -19,11 +19,9 @@ from web3.main import Web3
 
 
 def get_config(config_file: Optional[str] = None) -> Config:
-    config_file = config_file if config_file else os.getenv("CONFIG_FILE", "config.ini")
-
-    return (
-        Config(filename=config_file)
-        if config_file
+    return Config(
+        filename=config_file
+        if config_file is not None
         else os.getenv("CONFIG_FILE", "config.ini")
     )
 
--- a/ocean_provider/utils/basics.py
+++ b/ocean_provider/utils/basics.py
@@ -10,11 +10,11 @@ from ocean_lib.common.aquarius.aquarius import Aquarius
 from ocean_lib.models.data_token import DataToken
 from ocean_lib.ocean.util import get_web3_connection_provider
 from ocean_lib.web3_internal.wallet import Wallet
-from ocean_lib.web3_internal.web3_provider import Web3Provider
 from requests_testadapter import Resp
 
 import artifacts
 from ocean_provider.config import Config
+from web3.main import Web3
 
 
 def get_config():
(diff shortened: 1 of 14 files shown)


DMats:
I updated the description to say that this PR closes #174

tests/test_helpers.py
```diff
@@ -376,6 +377,10 @@ def send_order(client, ddo, datatoken, service, cons_wallet, expect_failure=Fals
         cons_wallet,
     )
     datatoken.verify_order_tx(
-        web3, tx_id, ddo.asset_id, service.index, amount, cons_wallet.address
+        tx_id, ddo.asset_id, service.index, amount, cons_wallet.address
     )
     return tx_id
+
+
+def get_web3():
```
DMats, line 385: This method is effectively a duplicate of `ocean_provider.utils.basics.get_web3()` - just use the latter instead.

ocean_provider/utils/util.py
```diff
@@ -329,7 +329,7 @@ def decode_from_data(data, key, dec_type="list"):
 def service_unavailable(error, context, custom_logger=None):
     text_items = []
     for key, value in context.items():
-        text_items.append(key + "=" + value)
+        text_items.append(key + "=" + str(value))
```
DMats, line 332: Why is this change necessary?

DMats (changes requested):

ocean_provider/utils/util.py
calina-c: Something broke byzantinely. I fixed it :)) Simple as that.

ocean_provider/utils/basics.py
```diff
@@ -17,14 +18,36 @@
...
-def get_config():
-    config_file = os.getenv("CONFIG_FILE", "config.ini")
-    return Config(filename=config_file)
+def get_config(config_file: Optional[str] = None) -> Config:
+    config_file = config_file if config_file else os.getenv("CONFIG_FILE", "config.ini")
 
+    return (
+        Config(filename=config_file)
+        if config_file
+        else os.getenv("CONFIG_FILE", "config.ini")
```
DMats, line 27: I just realized the code snippet I recommended in my last set of comments had a bug in it - the else returns the `CONFIG_FILE` string without building a `Config` instance :facepalm: So this code needs to be fixed.

Also, you're currently using 2 ternary operators but it can be done rather succinctly with just 1.

Here's my new suggestion - this time with no bugs! (hopefully, :crossed_fingers: )

```python
def get_config(config_file: Optional[str] = None) -> Config:
    return Config(
        filename=config_file
        if config_file is not None
        else os.getenv("CONFIG_FILE", "config.ini")
    )
```

tests/test_helpers.py
```diff
@@ -32,6 +32,7 @@
     get_datatoken_minter,
     get_artifacts_path,
```
DMats, line 33: Please remove unused imports including:

1. ExampleConfig
2. get_web3_connection_provider
3. get_artifacts_path
4. get_config

DMats (changes requested):

DMats (approved):

PR #179, opened 2021-07-12
Adds trusted publishers.

calina-c:
No issue to close yet. Will update that if necessary.

Btw, I used `trustedAlgorithmPublishers` as a key instead of `publisherTrustedAlgorithmPublishers` but I can change that. It seemed simpler.

### DIFF
--- a/ocean_provider/validation/algo.py
+++ b/ocean_provider/validation/algo.py
@@ -329,7 +329,9 @@ class InputItemValidator:
 
         return self.validate_usage()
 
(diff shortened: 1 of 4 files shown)


alexcos20 (changes requested):
let's use "publisherTrustedAlgorithmPublishers" pls
maybe it's cumbersome, but keeps things consistent, because the publisher will decide this list.

calina-c:
> let's use "publisherTrustedAlgorithmPublishers" pls
> maybe it's cumbersome, but keeps things consistent, because the publisher will decide this list.

Done!

alexcos20 (approved):

Issue #173, opened 2021-07-01
Provider seems buggy sometime, it is unresponsive for a couple of seconds


AnaLoznianu:
Sometimes we are getting this error 'Finding the service endpoints failed' when using Ocean.js directly with barge? Seems to happen at random without changing any code

```
Finding the service endpoints failed: AbortError: The user aborted a request.
    at abort (/home/jamie/Desktop/workshop/quickstart1/node_modules/node-fetch/lib/index.js:1418:16)
    at EventEmitter.abortAndFinalize (/home/jamie/Desktop/workshop/quickstart1/node_modules/node-fetch/lib/index.js:1433:4)
    at EventEmitter.emit (events.js:315:20)
    at AbortSignal.dispatchEvent (/home/jamie/Desktop/workshop/quickstart1/node_modules/node-abort-controller/index.js:28:23)
    at AbortController.abort (/home/jamie/Desktop/workshop/quickstart1/node_modules/node-abort-controller/index.js:40:17)
    at Timeout._onTimeout (/home/jamie/Desktop/workshop/quickstart1/node_modules/@oceanprotocol/lib/dist/node/lib.js:1:26814)
    at listOnTimeout (internal/timers.js:554:17)
    at processTimers (internal/timers.js:497:7) {
  type: 'aborted'
}
```
```
const run = async () => {
    const blob = `http://localhost:8030/api/v1/services/consume`;
    const ocean = await Ocean.getInstance(config);
    const accounts = await ocean.accounts.list();
    const alice = accounts[0];
    const bob = accounts[1];
    console.log('Alice account address:', alice.id)
    console.log('Bob account address:', bob.id)
    const datatoken = new DataTokens(
      contracts.DTFactory,
      factoryABI,
      datatokensABI,
      new Web3(urls.networkUrl)
    );
    const tokenAddress = await datatoken.create(
       blob, 
       alice.getId(),
       '10000000000',
       'AliceDT',
       'DTA');
    console.log(`Deployed datatoken address: ${tokenAddress}`);
    };
run();
```

alexcos20:
see https://github.com/oceanprotocol/ocean.js/issues/859 also

alexcos20:
so, the failing code in ocean.js is here: https://github.com/oceanprotocol/ocean.js/blob/666da8156fda8a3201b5b4cd600105017f65b275/src/provider/Provider.ts#L62

meaning that we are trying to get the endpoint of the provider, by calling the root endpoint.  and that is failing

[...]

PR #185, opened 2021-07-21
Issue173 add more log statements

DMats:
Towards #173 

Changes proposed in this PR:

- Add INFO log statements to the beginning of each route
  * Attempted DEBUG logs instead of INFO, but they don't appear, even when LOG_LEVEL=DEBUG.


mariacarmina (approved):
From my side, it looks fine.

PR #192, opened 2021-07-29
Renames config file env var to PROVIDER_CONFIG_FILE.

calina-c:
Related to https://github.com/oceanprotocol/multi-repo-issue/issues/103


DMats (changes requested):
Please change https://github.com/oceanprotocol/provider/blob/config-file-name/.env.example#L2 to `PROVIDER_CONFIG_FILE` as well.

Issue #203, opened 2021-09-01
Secure c2d output, get result API for non-signature 


soonhuat:
**Describe the bug**
With the new secure c2d output implementation done by @alexcos20 .
Suppose user only sign when they download actual compute result file. **But we still need to sign to get output list.**

Results object is the output file list.
If we **do not provide the signature** the results would be just empty []

If we **provide signature** to API https://provider-secure.rinkeby.dev.acentrik.io/api/v1/services/compute:
We will get "results": [
            {
                "filename": "heart_result_1.csv",
                "filesize": 241,
                "type": "output"
            },
            {
                "filename": "heart_result_2.csv",
                "filesize": 58,
                "type": "output"
            },
            {
                "filename": "algorithm.log",
                "filesize": 869,
                "type": "algorithmLog"
            }
        ]

**To Reproduce**
Just try to Get Results, and Download Output.

**Expected behavior**
When we don't provide signature we should still get 
"results": [
            {
                "filename": "heart_result_1.csv",
                "filesize": 241,
                "type": "output"
            },
            {
                "filename": "heart_result_2.csv",
                "filesize": 58,
                "type": "output"
            },
            {
                "filename": "algorithm.log",
                "filesize": 869,
                "type": "algorithmLog"
            }
        ]

**Potential fix and code line**
https://github.com/oceanprotocol/provider/blob/b49d605b920dd13c92d48cc217a1ccf2929d93b3/ocean_provider/routes/compute.py#L247
Maybe we can remove this line, so when request doesn't contains signature, respond still return with results object.
![image](https://user-images.githubusercontent.com/7642607/131614953-ece77616-1478-464b-8b93-31928f7c51dd.png)

calina-c:
@soonhuat can you try the GET /compute endpoint? That one doesn't require a signature. You can check on the job status as many times as possible using this endpoint, and only download the output when the job is done. The new endpoint (computeResult) outputs the binary result file, so it should require a signature, since the result should be secured. At least that is what I understand from Alex.

soonhuat:
@calina-c  ya, we are talking about /compute endpoint in this bug.
As above screenshot line 247, if no signature, the results list will be popped out from respond, so results list would for sure empty.
Isn't with feature of acquiring/download result is secure, the GET/showing results output metadata listing such what output 1, log 1, and it size should be available without secure?
So we view the list of output without signature, then download those with signature.
Else mean we need to sign to view the result list metadata, then sign again to download each of the output/log.
[...]

PR #205, opened 2021-09-03
Fix crash when signature is not sent on ComputeResult.

### LINKED ISSUE #204: Issue with /computeResult endpoint when no signature is provided
Need to isolate in a unit test and fix this.

calina-c:
Closes #204.

@alexcos20 I fixed the crash by bailing if the signature is not sent, but I see that #203 proposes an unsigned version of this endpoint. Can you advise? Should I work on a version where the signature is not required?

### DIFF
--- a/ocean_provider/validation/provider_requests.py
+++ b/ocean_provider/validation/provider_requests.py
@@ -187,7 +187,7 @@ class ComputeGetResult(CustomJsonRequest):
             "jobId": ["bail", "required"],
             "index": ["bail", "required"],
             "consumerAddress": ["bail", "required"],
-            "signature": ["required", "signature:consumerAddress,index,jobId"],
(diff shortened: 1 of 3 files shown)


alexcos20:
try to think from UX perspective.
I'm a data science guy that has 100 compute jobs started in the last week.  If I want to see my history, do I have to sign 100 times? I will never use that again in my life :)

So we need an endpoint that can accept a non signed request, but it will return non sensitive information

calina-c:
- remove signature from GET compute
- keep signature required for computeResult

calina-c:
Signature is already not required on the GET compute. It is optional. If we need to remove the option to add a signature on computeResult, altogether, please let me know. Otherwise this PR is as recommended.

PR #210, opened 2021-09-13
expose more compute limits

### LINKED ISSUE #209: Expose in root endpoint more compute limits
New op-service exposes some limits in the root endpoint:
```json
{"address":"0x6D39A833d1a6789AECa50dB85CB830bc08319F45","algoTimeLimit":"60","software":"Operator service","storageExpiry":2592000,"version":"1.0.0"}


Provider should expose that as well, in a computeLimits dict in the root endpoint

### PR DESCRIPTION
**Pull request by alexcos20**
Closes #209

### DIFF
--- a/ocean_provider/routes/compute.py
+++ b/ocean_provider/routes/compute.py
@@ -220,6 +220,7 @@ def computeStatus():
 
         _response = response.content
 
+        # The following sections is needed only to ensure back compat. It will be removed soon, make sure you are going to update your C2D backends
         signed_request = bool(data.get("signature"))
         if signed_request:
             owner = data.get("consumerAddress")
--- a/ocean_provider/utils/util.py
+++ b/ocean_provider/utils/util.py
@@ -203,7 +203,7 @@ def get_compute_info():
         compute_info = requests.get(get_config().operator_service_url).json()
         limits = {
             "algoTimeLimit": compute_info.get("algoTimeLimit"),
-            "storageExpiry": compute_info.get("storageExpiry")
+            "storageExpiry": compute_info.get("storageExpiry"),
         }
         compute_address = compute_info.get("address", None)
         return compute_address, limits
--- a/ocean_provider/utils/util.py
+++ b/ocean_provider/utils/util.py
@@ -207,6 +207,18 @@ def get_compute_address():
         return None
 
 
+def get_compute_limits():
+    try:
+        limits = dict()
+        compute_info = requests.get(get_config().operator_service_url).json()
+        limits["algoTimeLimit"] = compute_info.get("algoTimeLimit", None)
+        limits["storageExpiry"] = compute_info.get("storageExpiry", None)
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: ocean_provider/routes/compute.py [hunk 1]
```diff
@@ -220,6 +220,7 @@ def computeStatus():
 
         _response = response.content
 
+        # The following sections is needed only to ensure back compat. It will be removed soon, make sure you are going to update your C2D backends
```
calina-c, line 223: can you add this to the Release Notes when we upgrade the version? As a future deprecation?

ocean_provider/utils/util.py
```diff
@@ -198,13 +198,17 @@ def get_compute_result_endpoint():
     return get_config().operator_service_url + "/api/v1/operator/getResult"
 
 
-def get_compute_address():
+def get_compute_info():
     try:
+        limits = dict()
         compute_info = requests.get(get_config().operator_service_url).json()
-        return compute_info.get("address", None)
+        limits["algoTimeLimit"] = compute_info.get("algoTimeLimit", None)
```
calina-c, line 205: We have the util for filtering a dictionary https://github.com/oceanprotocol/provider/blob/b49d605b920dd13c92d48cc217a1ccf2929d93b3/ocean_provider/utils/util.py#L308

I think you can add a default to None parameter to it, and reuse it. I can do it if you're too busy.

ocean_provider/utils/util.py
```diff
@@ -198,13 +198,17 @@ def get_compute_result_endpoint():
...
+        limits["storageExpiry"] = compute_info.get("storageExpiry", None)
+        compute_address = compute_info.get("address", None)
+        return compute_address, limits
```
calina-c, line 208: I don't like the fact that we are returning a tuple, build from a dictionary. We can return the dictionary with "algoTimeLimit", "storageExpiry" and "address" and it would make more sense.

calina-c (changes requested):
Just a few changes. I'm happy to do them myself if necessary.

ocean_provider/utils/util.py
alexcos20: why should we do this?

ocean_provider/utils/util.py
alexcos20: we need a tuple, because that function is used in multiple places, and there is no point in returning a dict and then pop out values out of it based on needs

ocean_provider/utils/util.py
calina-c: I think it would make for a clearer code. An alternative would be to not declare dict separately and simply do:

```
limits = {
    "algoTimeLimit": compute_info.get("algoTimeLimit"),
    ...
}
```

to avoid declaring the dict() previously. Or simply return this dict value altogether, alongside the address.
Btw. there is no need for the `, None` part, since None is the default fallback.

PR #217, opened 2021-09-27
Upgrade to ocean.py 0.8.4

calina-c:


mariacarmina:
LGTM. Codeclimate shows only the duplicate parameters from the pytest fixtures.

AnaLoznianu:
Closing this, as we are removing ocean.py as a dependency for V4 integration

alexcos20 (approved):
lgtm

ocean_provider/config.py
```diff
@@ -56,6 +57,11 @@
         "resources",
     ],
     NAME_STORAGE_PATH: ["STORAGE_PATH", "Path to the local database file", "resources"],
+    NAME_BLOCK_CONFIRMATIONS: [
+        "BLOCK_CONFIRMATIONS",
```
DMats, line 61: I'm adding this as just a comment (instead of a change request) because this problem should go away once we remove the ocean.py dependency from the provider in https://github.com/oceanprotocol/provider/issues/218

Allowing provider `Config` and ocean.py `Config` to use the same `BLOCK_CONFIRMATIONS` envvar and `block_confirmations` config option creates unexpected coupling between the two code bases. It would be better to add a prefix to differentiate the two. `PROVIDER_BLOCK_CONFIRMATIONS` & `provider_block_confirmations`

calina-c: As I understand it, this is a temporary fix, where Alex bumped the ocean.py because he doesn't have time to review ocean.py removal right now.

alexcos20:
IPFS tests are failing, but that is a known issue.
Tested with latest ocean.js, all good, so I'm gonna merge this and release a new version

PR #219, opened 2021-09-28
Remove ocean.py dependency

calina-c:
Closes #218

PR #220, opened 2021-09-29
Fixed IPFS tests.

### LINKED ISSUE #214: Fix IPFS tests
Currently, the IPFS tests can randomly fail. Most likely, this happens because of bad connection, and it is tolerable since we run both PR and branch tests, so if one fails we can rely on the other. However, we should fix this and make sure IPFS tests work properly all the time.

Difficult to reproduce but we have to do this :)

mariacarmina:
Fixes #214  .

Changes proposed in this PR:

- add IPFS public [gateway()](http://ipfs.io) to the IPFS path.

### DIFF
--- a/ocean_provider/utils/test/test_util.py
+++ b/ocean_provider/utils/test/test_util.py
@@ -139,7 +139,7 @@ def test_build_download_response():
 
 def test_download_ipfs_file(client):
     cid = "QmQfpdcMWnLTXKKW9GPV7NgtEugghgD6HgzSF6gSrp2mL9"
-    url = f"ipfs://{cid}"
+    url = f"http://ipfs.io/ipfs/{cid}"
     download_url = get_download_url(url, None)
     requests_session = get_requests_session()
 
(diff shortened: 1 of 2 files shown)


mariacarmina:
Discussed with @calina-c and I will keep this PR as a draft until #219 will be merged.

ocean_provider/utils/test/test_util.py
```diff
@@ -139,7 +139,7 @@ class Dummy:
 
 def test_download_ipfs_file(client):
     cid = "QmQfpdcMWnLTXKKW9GPV7NgtEugghgD6HgzSF6gSrp2mL9"
-    url = f"ipfs://{cid}"
+    url = f"http://ipfs.io/ipfs/{cid}"
```
calina-c, line 142: this is not right. This is a test for a http-resolved ipfs URL. We want to test that an ipfs:// file is resolved and downloaded right.

calina-c (changes requested):

ocean_provider/utils/test/test_util.py
```diff
@@ -148,10 +148,11 @@ def test_download_ipfs_file(client):
 
     print(f"got ipfs download url: {download_url}")
     assert download_url and download_url.endswith(f"ipfs/{cid}")
-    response = build_download_response(
-        request, requests_session, download_url, download_url, None
-    )
-    assert response.data, f"got no data {response.data}"
+
+    response = requests_session.get(download_url, stream=True)
```
calina-c, line 152: it has to work with build_download_response, or it build_download_response has to be fixed. Please keepp tests on point. We want to test the ipfs download functionality, through provider.

calina-c (approved):

PR #261, opened 2021-11-29
Issue259 add compute algo service id param

### LINKED ISSUE #259: Add algoServiceId to compute requests
We need to add the algoServiceId to compute requests.
Please make sure it gets the same treatment as the serviceId (required)
Wherever we "guess" the compute service inside the flow, use the provided algoServiceId parameter instead.
Adjust tests to include this parameter.

See in https://github.com/oceanprotocol/provider/pull/258:
-> the remaining RBAC tests fail because on the "algos" array we are using the serviceId. We need to use algoServiceId instead (the change will be done in the RBACValidator class, not the tests)

maldo-ctrl:
Fixes #259 .

Changes proposed in this PR:

- Add required compute param `algorithmServiceId`
- Update `WorkflowValidator` to use new  `algorithmServiceId` param
- Update `RBACValidator` with new `algorithmServiceId` param
- Update tests


tests/test_RBAC.py
```diff
@@ -173,7 +174,7 @@ def test_compute_payload_without_additional_inputs(
     assert payload["dids"][0]["did"] == ddo.did
     assert payload["dids"][0]["serviceId"] == sa.id
     assert payload["algos"][0]["did"] == alg_ddo.did
-    assert payload["algos"][0]["serviceId"] == sa_compute.id
+    assert payload["algos"][0]["algorithmServiceId"] == sa_compute.id
```
maldo-ctrl, line 177: I have updated this tests too @calina-c . Is it OK?

ocean_provider/validation/algo.py
```diff
@@ -145,15 +146,7 @@ def _build_and_validate_algo(self, algo_data):
...
-                dt = get_datatoken_contract(self.web3, self.consumer_address)
-                tx_receipt = get_tx_receipt(self.web3, algorithm_tx_id)
-                event_logs = dt.events.OrderStarted().processReceipt(
-                    tx_receipt, errors=DISCARD
-                )
-
-                order_log = event_logs[0] if event_logs else None
-                algo_service_index = order_log.args.serviceIndex
-                self.algo_service = algo.get_service_by_index(algo_service_index)
+                self.algo_service = algo.get_service_by_id(algorithm_service_id)
```
calina-c, line 149: please keep the order check, we need to check that the service was ordered. Retrieve the algorithm service based on the provided argument, then check that the algorithm_tx_id indeed reflects that same algorithm Service. If it does not, add the error that the algo tx id and algo service id do not correspond. And please also add a test for this.

ocean_provider/validation/algo.py
```diff
@@ -266,6 +259,10 @@ def validate(self):
             self.error = "No serviceId in input item."
             return False
 
+        if not self.data.get("algorithmServiceId"):
```
calina-c, line 262: you can move this upward, when we are checking for the order log.

tests/test_RBAC.py
calina-c: nope, I think we should keep the structure, it's just the population that differs. See there is already "algos" in the key, so it should be {"algos": {"did": algo did, "serviceId": algo service id}}, but the key inside remains "serviceId" for the RBAC request

ocean_provider/validation/RBAC.py
```diff
@@ -60,7 +60,7 @@ def get_algos(self):
         return [
             {
                 "did": self.request["algorithmDid"],
-                "serviceId": self.request["serviceId"],
+                "algorithmServiceId": self.request["algorithmServiceId"],
```
calina-c, line 63: please keep the key as "serviceId" and the value as self.request["algorithmServiceId"]

calina-c (changes requested):
A few changes requested, but otherwise this looks neat!

ocean_provider/validation/algo.py
maldo-ctrl: The order check is performed here, if I'm not wrong: https://github.com/oceanprotocol/provider/blob/d1fafca4a1dbbe1af1d0a0a8abe3ec204be54ad9/ocean_provider/validation/algo.py#L164

calina-c (approved):
Amazing work! Happy to approve!

PR #264, opened 2021-12-02
Improve test algo coverage

### LINKED ISSUE #224: Enhance code coverage after urgent removal of ocean.py dependency
In #218 we tackled removal of ocean.py dependencies. Some of the code was moved directly from ocean.py, where it is already tested. Due to the urgency of this task, provider code coverage decreased, and we have to put it back where it was (~90%)

md0x:
Signed-off-by: Pablo Maldonado <pablo@oceanprotocol.com>

Fixes #224 .

Changes proposed in this PR:

- Improve test algorithm validation coverage

image


ocean_provider/validation/algo.py
```diff
@@ -270,6 +270,10 @@ def validate(self):
 
         self.service = self.asset.get_service_by_id(self.data["serviceId"])
 
+        if not self.service:
```
maldo-ctrl, line 273: Moved this check before `check_asset_consumable` to prevent key errors when accessing `self.service.service_endpoint`

calina-c:
I'm planning of turning this into unit tests. I can't stand the fact that we're doing a full publishing flow here everytime. We should only be doing that in integration tests, not for testing an internal validation class. Plus after https://github.com/oceanprotocol/provider/pull/309 the params will change and this PR fails the checks, so I'm docking this PR for now.

calina-c:
@maldo-ctrl lmk when you want to have a discussion. Algo validation tests are now unit tests. I can give you a quick run through to show you the gist of it, and you can add as much coverage as possible now, without slow-downs.

PR #265, opened 2021-12-02
Fix some todos.

calina-c:


DMats:
@alexcos20 @calina-c In the download endpoint in provider, do we still need to validate that a transfer is not used for another service?

alexcos20:
yes, the order event has a specific field, called serviceIndex.  Both should match

alexcos20 (approved):
We will make sure that services are a match in a new PR

calina-c:
@David please handle the pending TODOs in a separate PR. Alex answered your question but this should become a separate issue.

PR #280, opened 2021-12-06
Update `artifacts/install.sh` script

DMats:
Changes proposed in this PR:

- Remove the line from `artifacts/install.sh` that copies the `artifacts/addresses` which doesn't exist (before this PR was opened).
- Remove `artifacts/address.json` because it is provided by barge and the copy stored here is probably out of date.


calina-c:
Please put back install.sh. It's still useful to us locally. install-remote.sh is only useful in the CI/tox context, not in the virtual env locally. Since the entire artifacts folder is just a temporary helper for us, let's keep it as helpful as possible. :D

mariacarmina (approved):

calina-c (changes requested):
See comment :)

DMats:
@calina-c I added `install.sh` back in, but I removed the line that copies the `artifacts/addresses` directory which no longer exists (even before this PR). Please let me know if more changes are needed.

PR #289, opened 2021-12-09
Replicates asset urls endpoint in v4.

### LINKED ISSUE #282: Expose asset URL(s) (V4)
Right now, the asset url(s) are stored encrypted.
But what if you want to : 
 - add a new url
 - change a provider for a specific service
Then, you need the plain url.  Of course, we need to make sure that only NFT Owner has rights to access this endpoint.

Q: Do we need to add additional roles? Should metadataupdater be allowed as well ?

Input required:  did, serviceId , nonce , signature
Output:  json object

calina-c:
Closes #282.


calina-c:
Will mark as ready for review after the tests pass.

alexcos20 (approved):

ocean_provider/routes/consume.py
```diff
@@ -284,6 +287,86 @@ def download():
...
+    try:
+        asset = get_asset_from_metadatastore(get_metadata_url(), did)
+        service = asset.get_service_by_id(service_id)
+        assert service.type == "access"
+
+        dt_token = get_web3().eth.contract(
+            abi=ERC20Template.abi, address=service.datatoken_address
+        )
+
+        if not dt_token.caller.isMinter(publisher_address):
```
alexcos20, line 342: this is not right.
we need to check for nft owner.

alexcos20 (changes requested):

ocean_provider/routes/consume.py
calina-c: gotcha, will do.

PR #296, opened 2021-12-13
Lighten RBAC tests.

### LINKED ISSUE #295: Provider tests for RBAC should not do any publishing
In the tests/test_RBAC file, we are doing asset publishing. Please use mocks instead. We are basically reproducing the download and compute flow tests, just to get some ids. Integration tests take a lot of time and we would be better off testing just the RBAC integration, without doing full publishes.

If it is not possible to remove all publishing, we should at least limit it.

calina-c:
Closes #295

### DIFF
--- a/tests/test_RBAC.py
+++ b/tests/test_RBAC.py
@@ -3,8 +3,8 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 import copy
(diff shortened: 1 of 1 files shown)


DMats (approved):
LGTM.  Great work.

PR #304, opened 2021-12-20
Feature/v4 compute

alexcos20:
Since provider will support multiple storage type, we should not try to send the url, but instead send it as remote.
pod-configure will download it


ocean_provider/routes/consume.py
```diff
@@ -19,6 +19,7 @@
     get_provider_wallet,
     get_web3,
 )
+from ocean_provider.utils.address import isAddressMatch
```
calina-c, line 22: please see my comment on the other PR regarding `snake_case`.

ocean_provider/serializers.py
```diff
@@ -43,23 +48,14 @@ def serialize(self):
...
-            if userdata:
-                dict_template["remote"]["algouserdata"] = userdata
-
         dict_template["container"] = algo_asset.metadata["algorithm"]["container"]
+        dict_template["remote"] = {
+            "serviceEndpoint": self.algo_service.service_endpoint,
+            "txId": algorithm_tx_id,
+            "serviceId": self.algo_service.id,
+        }
+        # userdata = self.algo_data.get("algouserdata")
```
calina-c, line 57: why is this commented out? Should we remove it entirely? Otherwise please leave a comment indicating why it's here and commented.

calina-c (changes requested):

ocean_provider/serializers.py
alexcos20: fixed

PR #305, opened 2021-12-20
add provider fees

alexcos20:
For every initialize, provider will output an additional object, called providerFee.  Example:

```json
{
            "dataToken": "0x123",
            "nonce": 1,
            "computeAddress": "0x456",
            "providerFee": 
            {
                "providerFeeAddress": "0xe08A1dAe983BC701D05E492DB80e0144f8f4b909", 
               "providerFeeToken": "0x0000000000000000000000000000000000000000", 
               "providerFeeAmount": 0,
               "providerData": "0x7b2274696d656f7574223a307d",
               "v": 28,
               "r": "0x73f30a7d25109fc390c5a600e492ba569345ad9d70afc55d873085036adfe68b", 
               "s": "0x237598dd8d9b8f1e6e20ec2656e5951ccec717d1c310e8952238e5d991acfbf7"
         }
}

```

That object is going to be used by ocean.js/ocean.py to startOrder function, so we can always check the provider fees

-


mihaisc:
Are fields documented somewhere ? like `providerData` , `v`, `r`, `s`

alexcos20:
https://github.com/oceanprotocol/contracts/blob/v4main_postaudit/contracts/templates/ERC20Template.sol#L395-L401

alexcos20:
I will add more docs when the fee engine will be built.  Right now, it's just hardcoded to address zero.

ocean_provider/utils/address.py
```diff
@@ -28,3 +28,9 @@ def get_contract_address(
         for chain_addresses in address_json.values()
         if chain_addresses["chainId"] == chain_id
     )
+
+
+def isAddressMatch(address1: str, address2: str) -> bool:
```
calina-c, line 33: I think we already have this somewhere. Either way, you can `return Web3.toChecksumAddress(address1) == Web3.toChecksumAddress(address2)` directly, instead of doing "if True return True"

ocean_provider/utils/datatoken.py
```diff
@@ -49,6 +52,33 @@ def verify_order_tx(
     if tx_receipt.status == 0:
         raise AssertionError("order transaction failed.")
 
+    # check provider fees
+    datatoken_contract = get_datatoken_contract(web3, datatoken_address)
```
calina-c, line 56: this line is now dupplicated. Please remove it from line 82.

ocean_provider/utils/address.py
calina-c: also, please use `snake_case` in Python.

ocean_provider/utils/provider_fees.py
```diff
@@ -0,0 +1,57 @@
+from web3.main import Web3
```
calina-c, line 1: can you please do an isort of these imports?

ocean_provider/utils/provider_fees.py
```diff
@@ -0,0 +1,57 @@
...
+    message = Web3.solidityKeccak(
+        ["bytes", "address", "address", "uint256"],
+        [
+            Web3.toHex(Web3.toBytes(text=provider_data)),
+            provider_fee_address,
+            provider_fee_token,
+            provider_fee_amount,
+        ],
+    )
+    provider_wallet = get_provider_wallet()
```
calina-c, line 41: this line is duplicated (line 25 and 41). Does the value change?

calina-c (dismissed):
I left a few comments. I'd also like to see some testing on the new functionality, but if you are pressed for time please create a ticket and assign it to one of the team members.

alexcos20:
the new functionality is already tested , because startOrder (in the smartcontract) function is doing that.
Later, we will add a fee engine, so we can compute fees based on file size, compute duration, compute cpu no required, etc.

Right now, the fees are hardcoded to 0, because it does not matter

ocean_provider/utils/provider_fees.py
alexcos20: fixed

calina-c (approved):

PR #311, opened 2021-12-27
update ERC721 template

### LINKED ISSUE #310: Validate metadata publishing on chain
Update tests, because ERC721 setmetadata has changed

alexcos20:
Closes #310


alexcos20:
so, we hit a circular dependency.  
Aqua needs Provider (with updated code).  But in order to update Provider, we need Aqua with updated code.

So both PRs ([this ](https://github.com/oceanprotocol/aquarius/pull/656) and https://github.com/oceanprotocol/provider/pull/311) should be merged in the same time.

To test it , do the following:
 - clone provider, build your own image
 - clone aquarius, build your own image
 - update barge to use those images , instead of default aqua & provider
 - Run the tests
 - Approve the PR

DMats (approved):

PR #314, opened 2021-12-30
Pass through algocustomdata.

calina-c:


ocean_provider/serializers.py
```diff
@@ -57,7 +57,7 @@ def serialize(self):
                 "txId": algorithm_tx_id,
                 "serviceIndex": self.algo_service.index,
```
alexcos20, line 58: Add this please:
"algouserdata": self.algo_data.get("algouserdata", None)

alexcos20 (changes requested):

alexcos20 (approved):

PR #318, opened 2022-01-04
fixes on initialize & download

alexcos20:
A couple of fixes for errors encountered during ocean.js development

### DIFF
--- a/ocean_provider/routes/consume.py
+++ b/ocean_provider/routes/consume.py
@@ -169,7 +169,7 @@ def initialize():
 
         file_index = int(data.get("fileIndex", "-1"))
         # we check if the file is valid only if we have fileIndex
-        if file_index > 0:
+        if file_index > -1:
             url_object = get_service_files_list(service, provider_wallet)[file_index]
             url_valid, message = validate_url_object(url_object, service_id)
             if not url_valid:
--- a/ocean_provider/routes/consume.py
+++ b/ocean_provider/routes/consume.py
@@ -268,8 +268,11 @@ def download():
         if service.type != ServiceType.ACCESS and Web3.toChecksumAddress(
             consumer_address
         ) != Web3.toChecksumAddress(compute_address):
-            return jsonify(
-                error=f"Service with index={service_id} is not an access service."
+            return (
+                jsonify(
+                    error=f"Service with index={service_id} is not an access service."
+                ),
+                400,
             )
         logger.info("validate_order called from download endpoint.")
         _tx, _order_log, _transfer_log = validate_order(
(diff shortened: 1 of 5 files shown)


ocean_provider/routes/consume.py
```diff
@@ -167,25 +167,26 @@ def initialize():
...
-        download_url = get_download_url(url_object, app.config["PROVIDER_CONFIG_FILE"])
-        download_url = append_userdata(download_url, data)
-        valid, url_details = check_url_details(download_url)
-
-        if not valid:
-            logger.error(
-                f"Error: Asset URL not found or not available. \n"
-                f"Payload was: {data}",
-                exc_info=1,
+        file_index = int(data.get("fileIndex", "-1"))
```
calina-c, line 170: I would say just return an error message if file index is not ok, then un-indent the contents of the "if" statement

ocean_provider/routes/consume.py
```diff
@@ -276,8 +277,10 @@ def download():
         )
 
         file_index = int(data.get("fileIndex"))
-
-        url_object = get_service_files_list(service, provider_wallet)[file_index]
+        files_list = get_service_files_list(service, provider_wallet)
+        if file_index > len(files_list):
+            return jsonify(error=f"No such fileIndex")
```
calina-c, line 282: can you also add a 400 status code? And I would suggest a tighter condition on the if, e.g. if the file index is not found in the file list. Possibly it's `if file_index not in files_list` or `if file_index not in enum(files_list)`

ocean_provider/utils/url.py
```diff
@@ -201,7 +201,7 @@ def _get_result_from_url(url, with_checksum=False):
 
 
 def append_userdata(url, data):
-    userdata = data.get("userdata")
+    userdata = data.get("userdata", None)
```
calina-c, line 204: yes, sorry for missing that :)

calina-c (changes requested):

ocean_provider/routes/consume.py
alexcos20: this piece of code checks the validity of a file , as long as we have fileIndex defined.
If not, provider will return data for initialize, but without checking the file (maybe frontend already did that with fileinfo endpoint)

ocean_provider/routes/consume.py
alexcos20: agreed with 400.   but for tighter condition, it's enough.  files_list is an array, you cannot have files_list[0] and files_list[2] , but file_.list[1] is missing.

calina-c (approved):

PR #327, opened 2022-01-18
Adds env var for private provider and usage in RBAC.

### LINKED ISSUE #324: The Provider config should have the option to be set as public/private
A recent upgrade to the RBAC server has now set up a check to find if a user is allowed to use the particular provider. We can improve this process by setting an env within the provider that can be set as `public` or `private` - that way the check can be skipped when the provider is public (e.g. all users are automatically allowed to use it). This would simplify the process and avoid an unnecessary API call.

calina-c:
Closes #324.

### DIFF
--- a/ocean_provider/validation/RBAC.py
+++ b/ocean_provider/validation/RBAC.py
@@ -78,10 +78,12 @@ class RBACValidator:
         ]
 
     def build_payload(self):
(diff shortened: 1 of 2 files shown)


alexcos20 (changes requested):
pls add the ENV to the docs, so we can know about it

PR #335, opened 2022-01-25
Issue334 add more logging

### LINKED ISSUE #334: Add more logging to provider

DMats:
Fixes #334.

Changes proposed in this PR:

* Improve logging
  - Remove extra calls to `setup_logging()` - This seems to have fixed DEBUG logs when LOG_LEVEL=DEBUG
  - Add INFO logging to all responses from provider endpoints
  - Add ERROR logging to all error responses from provider endpoints
  - Add `log_incoming_request` which logs all incoming requests, even invalid ones
  - Update `error_response()` to always return json.  All endpoints now return JSON if errors occur (even if their successful responses don't return json)
  - Remove test_log.py. tests no longer relevant after logic changes in setup_logging

* Other improvements
  - Register unit and integration markers with pytest. eliminate warnings
  - Remove unnecessary `transferTxId` handling from `util.process_compute_request`
  - Update `computeResult` endpoint so that more of the logic is wrapped in the `try/catch` block.
  - Add execute permission to `artifacts/intall-remote.sh`


ocean_provider/utils/util.py
```diff
@@ -10,25 +10,22 @@
 from cgi import parse_header
 
 import requests
+from artifacts import ERC721Template
```
calina-c, line 13: I think all artifact imports should be after jsonsempai

DMats: you're right.  isort got me.

calina-c (approved):

PR #351, opened 2022-02-08
add compute envs

### LINKED ISSUE #352: Compute Envs
C2d now exports the list of environments:

```
curl -X GET "http://localhost:31000/api/v1/operator/enviroments" -H  "accept: application/json"
```

```
[
  {
    "id": "ocean-compute",
    "nCPU": 1,
    "cpuType": "",
    "nGPU": 0,
    "gpuType": "",
    "ramGB": 1,
    "diskGB": 1,
    "priceMin": 0,
    "desc": "",
    "currentJobs": 0,
    "maxJobs": 10,
    "consumerAddress": "0x6D39A833d1a6789AECa50dB85CB830bc08319F45",
    "storageExpiry": 0,
    "maxJobDuration": 60,
    "lastSeen": 1644301977.91135
  }
]
```

Provider should:
 - expose an API Endpoint for libs to fetch the c2d envs
- enforce enviroment selection during initialize (if compute) and startCompute
- use envs to compute provider Fees

alexcos20:
Closes #352 

Changes proposed in this PR:

- 
- 
-


ocean_provider/utils/util.py
```diff
@@ -9,13 +9,13 @@
...
 from urllib.parse import urljoin
-import werkzeug
 
 import requests
-from jsonsempai import magic  # noqa: F401
+import werkzeug
 from artifacts import ERC721Template
 from eth_account.signers.local import LocalAccount
 from flask import Response
+from jsonsempai import magic  # noqa: F401
```
calina-c, line 18: please move this before artifacts, otherwise ERC721Template may not import

ocean_provider/utils/util.py
```diff
@@ -305,3 +309,12 @@ def check_asset_consumable(asset, consumer_address, logger, custom_url=None):
     logger.error(message, exc_info=1)
 
     return False, message
+
+
+def check_environment_exists(envs, id):
+    """Checks if enironment with id exists in environments list."""
+    if envs and isinstance(envs, list):
```
calina-c, line 316: is there a more Pythonic way to do this? I don't like the triple imbrication. Why are we iterating with a range-like for? Can we check with [env for env in envs if env["id"] == id] and take the first item?

I suggest:
```
if not envs or not isinstance(envs, list):
     return False

matching_envs = [env for env in envs if env["id"] == id]

return len(matching_envs) > 0
```

tests/test_compute.py
```diff
@@ -428,4 +441,4 @@ def test_compute_environments(client):
     compute_envs_endpoint = BaseURLs.SERVICES_URL + "/computeEnvironments"
     response = client.get(compute_envs_endpoint)
 
-    assert "environments" in response.json
+    assert "id" in response.json[0]
```
calina-c, line 444: since we are using test compute_helpers, I think we can also assert some actual contents, not just the presence of the key in the response

calina-c (changes requested):

alexcos20:
do not merge, we need to update ocean.js/py.

PR #360, opened 2022-02-10
Fix computeEnvironment endpoint error handling

DMats:
Changes proposed in this PR:

- Fix computeEnvironment endpoint error handling. If exception occurs when calling c2d backend, let handle_error return 503, service unavailable.

### DIFF
--- a/ocean_provider/utils/provider_fees.py
+++ b/ocean_provider/utils/provider_fees.py
@@ -68,18 +68,15 @@ def get_provider_fees(
 
 def get_c2d_environments() -> List:
     standard_headers = {"Content-type": "application/json", "Connection": "close"}
-    try:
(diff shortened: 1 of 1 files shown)


calina-c:
isn't this exception treated one level upward of this?

DMats:
> isn't this exception treated one level upward of this?
@calina-c 
The problem is that the Exception is handled here in `get_c2d_environments` and ignored - by returning an empty list `[]`.  One layer up, it returns the empty list with a status code = 200, indicating success.  

My edits make it so the Exception is *not* handled here.  Instead, it will be caught by `handle_error`, which will then return a status code = 503, indicating service unavailable.

calina-c (approved):

PR #368, opened 2022-02-14
Remove --with-thegraph, now implicit

calina-c:


alexcos20 (approved):
ipfs is already started by default by barge, so lgtm

PR #419, opened 2022-04-13
Fix c2d address detection in download endpoint

### LINKED ISSUE #418: Wrong c2d address detection
On the download Endpoint, we have a piece of code which detects if the files is being downloaded by C2d.  If yes, will allow that download, even is the service is "compute" type:
https://github.com/oceanprotocol/provider/blob/v4main/ocean_provider/routes/consume.py#L276-L287

This check doesn't take into account the computeEnvs, so for sure, it will not work

md0x:
Signed-off-by: Pablo Maldonado <pablo@oceanprotocol.com>

Fixes #418  .

Changes proposed in this PR:

- Retrieve c2d environments and compare consumer address with the service id to allow or block download

### DIFF
--- a/ocean_provider/routes/consume.py
+++ b/ocean_provider/routes/consume.py
@@ -278,7 +278,7 @@ def download():
 
     is_c2d_consumer_address = bool(
         [
-            Web3.toChecksumAddress(env["consumerAddress"])
+            True
             for env in c2d_environments
             if service.id == env["id"]
             and Web3.toChecksumAddress(env["consumerAddress"])
--- a/ocean_provider/routes/consume.py
+++ b/ocean_provider/routes/consume.py
@@ -273,15 +273,20 @@ def download():
     asset = get_asset_from_metadatastore(get_metadata_url(), did)
     service = asset.get_service_by_id(service_id)
 
-    compute_address, compute_limits = (
-        get_compute_info() if os.getenv("OPERATOR_SERVICE_URL") else None,
-        None,
+    # allow our C2D to download a compute asset
+    c2d_environments = get_c2d_environments()
+
+    is_c2d_consumer_address = bool(
+        [
+            Web3.toChecksumAddress(env["consumerAddress"])
+            for env in c2d_environments
+            if service.id == env["id"]
+            and Web3.toChecksumAddress(env["consumerAddress"])
+            == Web3.toChecksumAddress(consumer_address)
+        ]
     )
 
-    # allow our C2D to download a compute asset
-    if service.type != ServiceType.ACCESS and Web3.toChecksumAddress(
-        consumer_address
-    ) != Web3.toChecksumAddress(compute_address):
+    if service.type != ServiceType.ACCESS and not is_c2d_consumer_address:
         return error_response(
             f"Service with index={service_id} is not an access service.", 400, logger
         )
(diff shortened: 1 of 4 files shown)


ocean_provider/routes/consume.py
```diff
@@ -273,15 +273,20 @@ def download():
...
-        get_compute_info() if os.getenv("OPERATOR_SERVICE_URL") else None,
-        None,
+    # allow our C2D to download a compute asset
+    c2d_environments = get_c2d_environments()
+
+    is_c2d_consumer_address = bool(
+        [
+            True
+            for env in c2d_environments
+            if service.id == env["id"]
```
alexcos20, line 283: why do have service.id here ?
we just need to compare addresses..

alexcos20 (changes requested):

tests/test_download.py
```diff
@@ -220,3 +222,117 @@ def test_download_multiple_files(client, publisher_wallet, consumer_wallet, web3
...
+        service.index,
+        get_provider_fees(asset.did, service, consumer_wallet.address, 0),
+        consumer_wallet,
+    )
+
+    nonce = str(datetime.utcnow().timestamp())
+    _msg = f"{asset.did}{nonce}"
+
+    # Consume using url index and auth token
+    # (let the provider do the decryption)
```
DMats, line 249: I think this comment is incorrect. There's no auth token here. And when you say `(let the provider do the decryption)` I think you mean, `(let the provider fetch on-chain ddo using tx_id)`

The provider always does the decryption, whether it fetches the ddo from chain, or its provided in args.

tests/test_download.py
```diff
@@ -220,3 +222,117 @@ def test_download_multiple_files(client, publisher_wallet, consumer_wallet, web3
...
+        "documentId": asset.did,
+        "serviceId": service.id,
+        "consumerAddress": consumer_wallet.address,
+        "signature": sign_message(_msg, consumer_wallet),
+        "transferTxId": tx_id,
+        "fileIndex": 0,
+        "nonce": nonce,
+    }
+
+    def other_service(*args, **kwargs):
```
DMats, line 260: `*args, **kwargs` are unused.

tests/test_download.py
```diff
@@ -220,3 +222,117 @@ def test_download_multiple_files(client, publisher_wallet, consumer_wallet, web3
...
+        new_service = copy.deepcopy(service)
+        new_service.__setattr__("type", "compute")
```
DMats, line 262: I think this could be more readable as `new_service.type = "compute"`.

tests/test_download.py
DMats: Or you could just remove the comment.

tests/test_download.py
```diff
@@ -220,3 +222,117 @@ def test_download_multiple_files(client, publisher_wallet, consumer_wallet, web3
...
+        return new_service
```
DMats, line 319: See above comments.

tests/test_download.py
```diff
@@ -220,3 +222,117 @@ def test_download_multiple_files(client, publisher_wallet, consumer_wallet, web3
...
+
+    with patch("ocean_provider.routes.consume.get_c2d_environments") as mock:
+        mock.return_value = [
+            {
+                "priceMin": 60,
```
DMats, line 268: Why is `priceMin` needed here?

DMats (changes requested):

DMats (approved):

alexcos20 (approved):

PR #421, opened 2022-04-13
Fix bug where to_wei didn't handle decimals correctly

### LINKED ISSUE #420: `to_wei` doesn't handle decimals correctly
The `to_wei` function was copied from ocean.py. See for context: https://github.com/oceanprotocol/ocean.py/issues/783

DMats:
Fixes #420

Changes proposed in this PR:

- Fix the bug where to_wei didn't handle decimals correctly
- Update test_currency.py
-

### DIFF
--- a/ocean_provider/utils/currency.py
+++ b/ocean_provider/utils/currency.py
@@ -1,8 +1,10 @@
 from decimal import ROUND_DOWN, Context, Decimal, localcontext
 from typing import Union
 
+from eth_utils.currency import units
+
 """The maximum uint256 value."""
-MAX_UINT256 = 2**256 - 1
+MAX_UINT256 = 2 ** 256 - 1
 
 """decimal.Context tuned to accomadate MAX_WEI.
 * precision=78 because there are 78 digits in MAX_WEI (MAX_UINT256).
@@ -29,20 +31,6 @@ MIN_ETHER = Decimal("0.000000000000000001")
 """The maximum possible token amount on Ethereum-compatible blockchains, denoted in ether"""
 MAX_ETHER = Decimal(MAX_WEI).scaleb(-18, context=ETHEREUM_DECIMAL_CONTEXT)
 
-UNITS = [
-    "wei",
-    "kwei",
-    "mwei",
-    "gwei",
-    "szabo",
-    "finney",
-    "ether",
-    "kether",
-    "mether",
-    "gether",
-    "tether",
-]
-
 
 def to_wei(amount_in_ether: Union[Decimal, str, int]) -> int:
     return parse_units(amount_in_ether, DECIMALS_18)
@@ -56,7 +44,7 @@ def parse_units(
     float input is purposfully not supported
     """
     num_decimals = (
-        UNITS.index(unit_name) * 3 if isinstance(unit_name, str) else unit_name
+        int(units[unit_name].log10()) if isinstance(unit_name, str) else unit_name
     )
 
     decimal_amount = normalize_and_validate_unit(amount, num_decimals)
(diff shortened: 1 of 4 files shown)


ocean_provider/utils/currency.py
```diff
@@ -31,29 +29,58 @@
 """The maximum possible token amount on Ethereum-compatible blockchains, denoted in ether"""
 MAX_ETHER = Decimal(MAX_WEI).scaleb(-18, context=ETHEREUM_DECIMAL_CONTEXT)
 
+UNITS = [
```
calina-c, line 32: I'm not a fan of adding these here. Is there some library constant we ca use instead?

ocean_provider/utils/test/test_currency.py
```diff
@@ -52,13 +60,37 @@ def test_to_wei():
...
+    assert parse_units("0", USDT_DECIMALS) == 0
+    assert parse_units("0.123456789123456789", USDT_DECIMALS) == 123456
+    assert parse_units("1.123456789123456789", USDT_DECIMALS) == 1_123456
+    assert parse_units("5278.02", USDT_DECIMALS) == 5278_020000
+    assert parse_units(MIN_USDT, USDT_DECIMALS) == MIN_WEI
+    assert parse_units(MAX_USDT, USDT_DECIMALS) == MAX_WEI
+
+    with pytest.raises(ValueError):
+        # Use ETHEREUM_DECIMAL_CONTEXT when performing arithmetic on MAX_USDT
+        with localcontext(ETHEREUM_DECIMAL_CONTEXT):
```
calina-c, line 76: I think the test is more relevant if we move the localcontext part outside (switching the `with`s), to scope out the ValueError. i.e. to outrule the possibility that localcontext itself throws the ValueError. Same goes for a few lines now, for the same usage.

calina-c (changes requested):

ocean_provider/utils/currency.py
DMats: eth-utils currency.py has a `units` dictionary that maps names to their corresponding `Decimal` value. https://github.com/ethereum/eth-utils/blob/d8b5aa197c3bf6d9d23b677556373f44a6aa3320/eth_utils/units.py#L6-L30

Using `units` to calculate the exponent requires logarithms (which admittedly, I'm rusty at) so at the time, just copying the `UNITS` list from ethers.js was faster. 

But now that I'm revisiting, I took the time to figure out the logarithm logic which will allow me to remove the `UNITS` list.

```python
    num_decimals = (
        int(units[unit_name].log10()) if isinstance(unit_name, str) else unit_name
    )
```

NOTE:  I chose not to use `dict.get()` because I think this method should throw an exception if the user passes in an unsupported unit name.

The corresponding ocean.py PR has already been merged: https://github.com/oceanprotocol/ocean.py/pull/784 so I'll create a new PR in ocean.py with this fix.

ocean_provider/utils/test/test_currency.py
DMats: Sure this makes sense to me.  I'll make sure to include this change in the new ocean.py PR I mentioned in my comment above.

DMats:
The tests seem to be failing due to errors unrelated to this PR. Did the `get_c2d_environments` response format change?

Example:
```python
        environments = get_c2d_environments()
        ddo, tx_id, alg_ddo, _ = build_and_send_ddo_with_compute_service(
            client,
            publisher_wallet,
            consumer_wallet,
            False,
            None,
>           environments[0]["consumerAddress"],
        )
E       IndexError: list index out of range

tests/test_compute.py:115: IndexError
```

calina-c:
See https://github.com/oceanprotocol/provider/pull/424/commits/33c8c4890f8ef8e99bf868d36b0ef331ff316321

calina-c (approved):
Seems perfect. Let's see if we merge https://github.com/oceanprotocol/provider/pull/424 first. Otherwise I think you can cherry pick just the comment I pasted above to fix the tests.

Issue #381, opened 2022-02-25
OrderExecuted


alexcos20:
What happends after a startOrder ?
Consumer will reach out to provider, trying to download file or start a computeJob
And he succeeds.

Nevertheless, we will claim that he never downloaded or start a computeJob, which leads to claim resolving.
How to fix that ?

Proposal: Have a set of ENV vars, that control how Provider creates the proof:
  - USE_CHAIN_PROOF:    if true, then Provider calls ERC20. OrderExecuted (ERC20.https://github.com/oceanprotocol/contracts/blob/v4main_postaudit/contracts/templates/ERC20Template.sol#L1081)
  - USE_HTTP_PROOF:  calls a HTTP endpoint instead of calling ERC20.OrderExecuted (POST query, body description below)

Parameters: 
For any kind of proof, we need the following:
 - orderTxId,
 - providerSignature,  (Provider signs providerData using it's own private key)
 - consumerData  (Consumer parameters that formed the signature message that Provider received)
 - consumerSignature  (Consumer signature is passthrough)
 - consumerAddress
 - providerData:     
 - - For download, we need a JSON object that holds:   DID, serviceIndex, fileIndex, downloadedBytes
 - - For compute, we need a JSON object that holds:   dataset DID, dataset serviceIndex, algo DID, algo serviceIndex, additionalDatasets (DIDs & serviceIndex)

calina-c:
Moving to backlog for now, signatures are incompatible and we need to table this for now

calina-c:
This is done except for the downloadedBytes part.

mihaisc:
@alexcos20 status on this? can we close it?

PR #424, opened 2022-04-18
Adds proofs, pending downloaded bytes

calina-c:
Work on #381

### DIFF
--- a/ocean_provider/utils/proof.py
+++ b/ocean_provider/utils/proof.py
@@ -58,4 +58,6 @@ def send_proof(
         {"from": provider_wallet.address, "gasPrice": int(web3.eth.gas_price * 1.1)}
     )
 
-    sign_send_and_wait_for_receipt(web3, tx, provider_wallet)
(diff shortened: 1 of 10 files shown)


ocean_provider/utils/proof.py
```diff
@@ -0,0 +1,63 @@
...
+        Web3.toBytes(text=provider_data),
+        provider_signature,
+        consumer_message,
+        consumer_signature,
+        consumer_address,
+    ).buildTransaction(
+        {"from": provider_wallet.address, "gasPrice": int(web3.eth.gas_price * 1.1)}
+    )
+
+    transaction_id, _ = sign_send_and_wait_for_receipt(web3, tx, provider_wallet)
```
alexcos20, line 61: Can we just send the tx and not wait for receipt?
It's going to block the thread and anyway, we are not having any fallback code (like increase the gas, etc)

alexcos20 (changes requested):

alexcos20 (approved):
lgtm

PR #426, opened 2022-04-20
Check deployer factory in decrypt.

### LINKED ISSUE #422: Decrypt: Make sure that NFT is created by our Factory
To increase security, when decrypting a DDO, always make sure that NFT is created by our Factory.

calina-c:
Closes #422

### DIFF
--- a/ocean_provider/routes/decrypt.py
+++ b/ocean_provider/routes/decrypt.py
@@ -8,7 +8,6 @@ from hashlib import sha256
 from typing import Optional, Tuple
 
 from eth_typing.encoding import HexStr
-from eth_typing.evm import HexAddress
 from flask import Response, request
 from flask_sieve import validate
 from ocean_provider.requests_session import get_requests_session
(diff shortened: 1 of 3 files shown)


ocean_provider/routes/decrypt.py
```diff
@@ -72,23 +72,27 @@ def _decrypt(
...
     logger.info(f"authorized_decrypters = {authorized_decrypters}")
     if authorized_decrypters and decrypter_address not in authorized_decrypters:
-        return error_response(f"Decrypter not authorized", 403, logger)
+        return error_response("Decrypter not authorized", 403, logger)
+
+    factory_contract = get_data_nft_factory_contract(web3)
+    if factory_contract.caller.erc721List(data_nft_address) != data_nft_address:
+        return error_response(
+            "Asset not deployed by the data NFT factory.", 400, logger
+        )
```
DMats, line 87: Let's add this as a function in `data_nft_factory.py`.

```python
def is_nft_deployed_from_factory(nft_address: str) -> bool:
    """Check if NFT is deployed from the factory."""
    data_nft_factory = get_data_nft_factory_contract(web3)
    return data_nft_factory.caller.erc721List(nft_address) == nft_address
```

Then the usage here would become

```python
if not is_nft_deployed_from_factory(data_nft_address):
        return error_response(
            "Asset not deployed by the data NFT factory.", 400, logger
        )
```

Also, as a side note, this comment showed me we're missing this method (and the corresponding one for datatokens) in ocean.py

DMats (changes requested):

DMats (approved):

PR #428, opened 2022-04-22
Separate initialize

### LINKED ISSUE #425: Provider fees for compute
Right now, if I will start a compute job with one algo and one dataset, I will pay provider fees twice.

Proposal:
 - in https://github.com/oceanprotocol/provider/blob/22b2a155f0ea9e27a8c4f8c6a80598ca88337b28/ocean_provider/utils/provider_fees.py#L36, make sure we charge providerFees for computeJobs only for algo did.
For computeStart, make sure env passed as argument(https://github.com/oceanprotocol/provider/blob/v4main/ocean_provider/routes/compute.py#L231) is a match

calina-c:
Works on https://github.com/oceanprotocol/multi-repo-issue/issues/146.
Closes #425.

- [x] after merging the docstrings branch, add docstring to initializeCompute and check that initialize docstring is still correct.


alexcos20:
1. how does the result looks like when I will call initializeCompute with one algo DID and one dataset DID, no validOrders on any of those ?   I will pay 2 providerFees  -> that is incorrect
2. how does the result looks like when I will call initializeCompute with one algo DID (with a valid orderID and valid providerFees) and one dataset DID, with no validOrders on any of those ?   I already paid providerFee for algo, and although it's still valid, I will pay again providerFee for datast  -> that is incorrect

ocean_provider/utils/provider_fees.py
```diff
@@ -77,6 +80,58 @@ def get_provider_fees(
     return provider_fee
 
 
+def get_provider_fees_or_remote(
```
alexcos20, line 83: this function is atomicly executed for every input -> leads to user paying providerFee multiple times.
See my previous comment

alexcos20 (changes requested):

alexcos20 (approved):

PR #437, opened 2022-05-03
Remove v3 local file adapter.

calina-c:


DMats (approved):
I ran the ocean.py test suite against these changes and everything passed. Visually, I can see that the `LocalFileAdapter` is removed, so I believe the issue should be fixed and an explicit test is not required. Thus, I am approving.

PR #441, opened 2022-05-10
Fix requirement for ADDRESS_FILE.

calina-c:


alexcos20 (commented):
I don't think it will work :)
It will always return addresses from imported contracts

alexcos20 (approved):

Issue #442, opened 2022-05-10
Make sure that order timeout is checked on consumption


alexcos20:
When startOrder is emitted (https://github.com/oceanprotocol/contracts/blob/v4main/contracts/templates/ERC20Template.sol#L535-L543), the timestamp is recorded.

When checking if the order is valid, provider should fetch the timeout from the ddo.services (see https://docs.oceanprotocol.com/concepts/did-ddo/#services)  and make sure that consumption is no longer authorized if that timeout has passed

This change needs to go into:
- [x] V3 provider #444
- [x] V4 provider #443 and #445

DMats:
Fix should probably go into `verify_order_tx`

```
// Pseudocode
if current_timestamp - order_event_timestamp < timeout:
    reject download request
```

DMats:
@calina-c requested that I add a test in v4 that shows `initializeCompute` also works with the new timeout logic.

PR #444, opened 2022-05-11
Issue442 enforce order timeout v3

DMats:
Towards #442.

Changes proposed in this PR:

- Enforce order timeout when consuming an asset (download or c2d). If timeout == 0, order valid forever. else, reject request if current timestamp - order timestamp > timeout.
- Add `test_download_timeout`
- Ensure that errors in `validate_order_tx` result in 400 error responses (previously resulted in 503)
- Move `get_metadata_url` and `get_datatoken_minter` to fix circular import issues
- Add `service_timeout` argument to `verify_order_tx`
- Add `timeout` argument to the `get_dataset_ddo_with_access_service` test helper method
- Remove `verify_order_tx` from the `send_order` test helper method, unnecessary.

### DIFF
--- a/ocean_provider/utils/datatoken.py
+++ b/ocean_provider/utils/datatoken.py
@@ -4,12 +4,9 @@ from jsonsempai import magic  # noqa: F401
 from artifacts import DataTokenTemplate
 from eth_utils import remove_0x_prefix
 from hexbytes import HexBytes
-from ocean_provider.utils.basics import (
-    get_asset_from_metadatastore,
-    get_metadata_url,
-    get_web3,
-)
+from ocean_provider.utils.basics import get_web3
 from ocean_provider.utils.currency import to_wei
+from ocean_provider.utils.services import Service
 from web3.logs import DISCARD
 from websockets import ConnectionClosed
 
@@ -49,7 +46,7 @@ def mint(web3, contract, receiver_address, amount, minter_wallet):
 
 
 def verify_order_tx(
-    web3, contract, tx_id: str, did: str, service_id, amount, sender: str
+    web3, contract, tx_id: str, did: str, service: Service, amount, sender: str
 ):
     try:
         tx_receipt = get_tx_receipt(web3, tx_id)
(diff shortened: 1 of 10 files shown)


DMats:
I believe the core content of this PR is ready for review. @calina-c @alexcos20   The `test_compute` is the only failing test, and it seems to be caused by an error in the c2d backend. I can look into it more deeply on Monday.
https://github.com/oceanprotocol/provider/runs/6429276995?check_suite_focus=true#step:7:194

ocean_provider/utils/datatoken.py
```diff
@@ -67,14 +82,31 @@ def verify_order_tx(
...
 
+    # Check if order expired. timeout == 0 means order is valid forever
+    service_timeout = service.main["timeout"]
+    timestamp_now = datetime.utcnow().timestamp()
+    timestamp_delta = timestamp_now - order_log.args.timestamp
+    logger.debug(
+        f"verify_order_tx: service timeout = {service_timeout}, timestamp delta = {timestamp_delta}"
+    )
+    if service_timeout != 0:
+        if timestamp_delta > service_timeout:
```
calina-c, line 101: I think if+if can be transformed to an "and"

ocean_provider/utils/util.py
```diff
@@ -218,9 +215,9 @@ def get_compute_info():
         return None, None
 
 
-def validate_order(web3, sender, token_address, num_tokens, tx_id, did, service_id):
+def validate_order(web3, sender, token_address, num_tokens, tx_id, did, service):
     logger.debug(
-        f"validate_order: did={did}, service_id={service_id}, tx_id={tx_id}, "
+        f"validate_order: did={did}, service_id={service.index}, tx_id={tx_id}, "
```
calina-c, line 220: service id should be kept for debugging, even if we are operating with the index on chain.

calina-c (changes requested):
I assume the tests are failing due to a mixup between the service id and the service index. Why are we doing this change now, on v3? I think we should only add the timeout-related fix and not change the structure of v3. Furthermore, I think in v3 we ARE using the index directly everywhere, and the change for using the id was only introduced in v4 (but I'm not entirely sure). I would suggest keeping the service-related parameters as they were.

ocean_provider/utils/util.py
DMats: I don't think there was a "service id" in v3.  In v3, `id` and `index` referred to the same thing.

DMats:
> I assume the tests are failing due to a mixup between the service id and the service index. Why are we doing this change now, on v3? I think we should only add the timeout-related fix and not change the structure of v3. Furthermore, I think in v3 we ARE using the index directly everywhere, and the change for using the id was only introduced in v4 (but I'm not entirely sure). I would suggest keeping the service-related parameters as they were.

Okay.  I'll try reverting this change to maintain the `service_id` argument. I'll instead add a new `service_timeout` argument.

ocean_provider/utils/datatoken.py
DMats: Done.

DMats:
> I assume the tests are failing due to a mixup between the service id and the service index. 

@calina-c I reverted the changes to the `service_id` argument but the `test_compute` test is failing in the same way as before. Thus, I think the above assumption is incorrect and the root cause must be something else.

DMats:
@calina-c @md00ux @alexcos20 
I believe the only remaining CI error (linked below) will be fixed once https://github.com/oceanprotocol/pod-publishing/pull/34 is merged and deployed.

https://github.com/oceanprotocol/provider/runs/6455098154?check_suite_focus=true#step:7:194

ocean_provider/utils/util.py
DMats: I originally reverted the changes to the `service_id` and added the `service_timeout` argument because we thought the previous changes were causing errors.  We now know the errors were from an external root cause.

I briefly considered "un-reverting" the changes to `service_id` again, but decided that I actually like the new `service_timeout` argument better.  It minimizes the data passed in - no need to pass the whole Service.

PR #445, opened 2022-05-12
Add initialize compute timeout tests

DMats:
Towards #442

Changes proposed in this PR:

- Test initialize compute, reuse order but the order is expired
- Improve debug log inside `verify_order_tx`
- Add `timeout` argument to `get_compute_service`, `get_compute_service_no_rawalgo`, and `build_custom_services`


DMats:
@calina-c Thanks for the comment.  I've updated the initialize compute test to use a loop like the ones in the commit you linked. Please review again when you get a chance.

calina-c (approved):

PR #450, opened 2022-05-17
fix #447

### LINKED ISSUE #447: Provider:  Allow 3rd party C2D providers
In order to allow 3rd party C2D providers, we need to stop exposing URL for assets served by the current provider, and always send the provider download link  (as we are doing for remote assets)

alexcos20:
Closes #447

### DIFF
--- a/ocean_provider/validation/algo.py
+++ b/ocean_provider/validation/algo.py
@@ -343,24 +343,18 @@ class InputItemValidator:
             if not self.validate_algo():
                 return False
 
-        if asset_urls:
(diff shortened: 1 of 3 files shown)


ocean_provider/validation/algo.py
```diff
@@ -343,24 +343,18 @@ def validate(self):
             if not self.validate_algo():
                 return False
 
-        if asset_urls:
```
calina-c, line 346: can you please check whether asset_urls is still used anywhere? If it is not, please remove the resolve.

calina-c (approved):

PR #466, opened 2022-05-31
use validUntil in c2d payload

### LINKED ISSUE #465: C2d job cutoff
When calling initializeCompute, user will specify the validUntil parameter  (I want to use c2d resources from now until 01 Jun 2022 11:34). And provider calculates the necessary providerFee which needs to be paid (if any)

When calling startCompute, we are sending maxSeconds  (https://github.com/oceanprotocol/provider/blob/v4main/ocean_provider/routes/compute.py#L454) which is generated in https://github.com/oceanprotocol/provider/blob/v4main/ocean_provider/routes/compute.py#L441-L443.

But validator does not take into account the providerFees, but only assets.timeout, which is wrong.

Proposal:
 - add "validUntil" object instead of maxSeconds in https://github.com/oceanprotocol/provider/blob/v4main/ocean_provider/routes/compute.py#L454
 - validUntil should be the maximum providerFee.validUntil from all datasets + algo (if any)  (always expressed as UTC timestamp)

Logic:
 - I want to  buy dataset D , with timeout 1 week
 - I want to  buy algo A, with timeout 1 month
 - Assume that current timestamp is 1th June 2022, 19:00:00
 - User will call initializeCompute, using env "e" and no txid for both dataset and algo, and wants 10 minutes of computeResources  -> providerFee.validUntil = "1th June 2022, 19:10:00".   
 - Provider Response:  order algo A + pay providerFee(feeAmount >0, validUntil= "1th June 2022, 19:10:00" and order dataset + pay 0 providerFee (feeAmount = 0, because fee it's already payed for algo A)
 - User orders dataset D by paying DTD + providerFee and algo A by paying DTA(no providerFees)
 - When user calls startCompute, Provider will detect providerFee from dataset D (0 fee) and providerFee from algo A (it has amount, token, validUntil = "1th June 2022, 19:10:00"
 - send "1th June 2022, 19:10:00" converted to UTC timestamp to C2D  (as payload.validUntil), so C2d can cut off the job if it's still running past "1th June 2022, 19:10:00"
 
In 2 days, user wants to start a new c2d job, using same dataset + algo:
 - Assume that current timestamp is 3th June 2022, 10:00:00
 - User will call initializeCompute, using env "e" and txidD + txidA  (user already has validOrders), and wants 25 minutes of computeResources  -> providerFee.validUntil = "3th June 2022, 10:25:00".   
 - Provider Response:  reuse order algo A + pay providerFee(feeAmount >0, validUntil= "3th June 2022, 10:25:00" and use txid of order dataset, because it's still valid
 - User reorders dataset D by paying only the new providerFee
 - When user calls startCompute, Provider will detect providerFee from dataset D (0 fee) and the new providerFee from algo A (it has amount, token, validUntil = "3th June 2022, 10:25:00"
 - send "3th June 2022, 10:25:00" converted to UTC timestamp to C2D  (as payload.validUntil), so C2d can cut off the job if it's still running past "3th June 2022, 10:25:00"

### LINKED ISSUE #471: PROVIDER_FEE_TOKEN default - use Ocean Address
If there is no PROVIDER_FEE_TOKEN env, use Ocean address (instead of Zero_Address)

alexcos20:
Closes #465
Closes #471

### DIFF
--- a/ocean_provider/utils/address.py
+++ b/ocean_provider/utils/address.py
@@ -35,9 +35,7 @@ def get_contract_address(
 
 def get_provider_fee_token(chain_id):
     fee_token = os.environ.get("PROVIDER_FEE_TOKEN", get_ocean_address(chain_id))
-    if not fee_token:
-        fee_token = BLACK_HOLE_ADDRESS
-    return fee_token
+    return fee_token if fee_token else BLACK_HOLE_ADDRESS
 
 
 def get_ocean_address(chain_id):
--- a/ocean_provider/utils/address.py
+++ b/ocean_provider/utils/address.py
@@ -3,6 +3,7 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 import json
+import os
 from pathlib import Path
 from typing import Any, Dict, Union
 from eth_typing.evm import HexAddress
--- a/ocean_provider/utils/address.py
+++ b/ocean_provider/utils/address.py
@@ -9,6 +9,7 @@ from typing import Any, Dict, Union
 from eth_typing.evm import HexAddress
 from ocean_provider.utils.basics import get_config
 
+BLACK_HOLE_ADDRESS = "0x0000000000000000000000000000000000000000"
 
 def get_address_json(address_path: Union[str, Path]) -> Dict[str, Any]:
     """Return the json object of all Ocean contract addresses on all chains."""
@@ -34,7 +35,7 @@ def get_contract_address(
 def get_provider_fee_token(chain_id):
     fee_token = os.environ.get("PROVIDER_FEE_TOKEN", get_ocean_address(chain_id))
     if not fee_token:
-        return "0x0000000000000000000000000000000000000000"
+        return BLACK_HOLE_ADDRESS
 
 
 def get_ocean_address(chain_id):
--- a/ocean_provider/utils/address.py
+++ b/ocean_provider/utils/address.py
@@ -2,7 +2,6 @@
 # Copyright 2021 Ocean Protocol Foundation
 # SPDX-License-Identifier: Apache-2.0
 #
-from itertools import chain
 import json
 from pathlib import Path
 from typing import Any, Dict, Union
--- a/ocean_provider/utils/address.py
+++ b/ocean_provider/utils/address.py
@@ -11,6 +11,7 @@ from ocean_provider.utils.basics import get_config
 
 BLACK_HOLE_ADDRESS = "0x0000000000000000000000000000000000000000"
 
+
 def get_address_json(address_path: Union[str, Path]) -> Dict[str, Any]:
     """Return the json object of all Ocean contract addresses on all chains."""
     if isinstance(address_path, str):
@@ -25,6 +26,9 @@ def get_contract_address(
 ) -> HexAddress:
     """Return the contract address with the given name and chain id"""
     address_json = get_address_json(address_path)
+    print(address_json)
+    print(contract_name)
+    print(chain_id)
     return next(
         chain_addresses[contract_name]
         for chain_addresses in address_json.values()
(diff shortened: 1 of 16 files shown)


ocean_provider/utils/address.py
```diff
@@ -27,3 +29,13 @@ def get_contract_address(
         for chain_addresses in address_json.values()
         if chain_addresses["chainId"] == chain_id
     )
+
+
+def get_provider_fee_token(chain_id):
+    fee_token = os.environ.get("PROVIDER_FEE_TOKEN", get_ocean_address(chain_id))
+    if not fee_token:
+        return "0x0000000000000000000000000000000000000000"
```
calina-c, line 37: there's a constant for ZERO_ADDRESS somewhere.

calina-c (approved):

ocean_provider/utils/address.py
```diff
@@ -27,3 +31,14 @@ def get_contract_address(
         for chain_addresses in address_json.values()
         if chain_addresses["chainId"] == chain_id
     )
+
+
+def get_provider_fee_token(chain_id):
+    fee_token = os.environ.get("PROVIDER_FEE_TOKEN", get_ocean_address(chain_id))
+    if not fee_token:
```
calina-c, line 38: i suggest using the ternary: `return fee_token if fee_token else BLACK_HOLE_ADDRESS`.

ocean_provider/utils/provider_fees.py
```diff
@@ -174,7 +173,7 @@ def get_provider_fee_amount(valid_until, compute_env, web3, provider_fee_token):
     if provider_fee_token == "0x0000000000000000000000000000000000000000":
         return 0
 
-    provider_fee_amount = float(seconds * env["priceMin"] / 60)
+    provider_fee_amount = float(seconds * float(env["priceMin"]) / 60)
```
calina-c, line 176: I also did this on the other branch 👍

tests/test_compute.py
```diff
@@ -276,14 +281,16 @@ def test_compute(client, publisher_wallet, consumer_wallet):
 
 @pytest.mark.integration
 def test_compute_diff_provider(client, publisher_wallet, consumer_wallet):
-    environments = get_c2d_environments()
+    environment = get_free_c2d_env()
```
calina-c, line 284: can we add this as a fixture in conftest.py instead?

ocean_provider/validation/algo.py
```diff
@@ -112,11 +113,10 @@ def validate_input(self, index=0):
             valid_until_list.append(self.algo_valid_until)
             provider_fee_amounts.append(self.algo_fee_amount)
 
-        self.valid_until = min(valid_until_list)
+        self.valid_until = max(valid_until_list)
+
+        provider_fee_token = get_provider_fee_token(self.web3.chain_id)
```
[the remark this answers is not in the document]
alexcos20: on next line? :)
```
required_provider_fee = get_provider_fee_amount(
            self.valid_until,
            self.data.get("environment"),
            self.web3,
            provider_fee_token,
        )
```

tests/test_compute.py
alexcos20: feel free to do it :)

ocean_provider/utils/address.py
alexcos20: coming soon

tests/test_helpers.py
```diff
@@ -477,3 +479,11 @@ def build_custom_services(
 def get_first_service_by_type(asset, service_type: ServiceType) -> Service:
     """Return the first Service with the given ServiceType."""
     return next((service for service in asset.services if service.type == service_type))
+
+
+def get_free_c2d_env():
+    environments = get_c2d_environments()
+    for env in environments:
```
calina-c, line 486: I would sugggest a comprehension, i.e:
`free_envs = [env for env in get_c2d_environments() if float(env["priceMin"]) == 0`
`return free_envs[0] if free_envs else None`

please check the conversion as well, I don't think the priceMin will be an int/float, but probably a string, so we need to convert it.

ocean_provider/validation/algo.py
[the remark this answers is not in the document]
calina-c: yep, sorry, did not see that

calina-c (approved):

PR #473, opened 2022-06-03
Duration

### LINKED ISSUE #472: ProviderFee validity
Our current setup works like:
 - let's assume that current timestamp is '3June2022 07:00:00'
 - user wants to run a c2d job for 10 minutes
 - user will call initializeCompute, with a validUntil = '3June2022 07:10:00'  (absolute timestamp)
 - user receives instructions how to order dataset & algo and how to pay providerFees
 - user generates startOrder transaction(s)
 - once transactions are confirmed, user calls startCompute, with the same validUntil

But,
What's going to happend if startOrder transactions gets confirmed 30 mins later, due to congested network?
User paid the fees, but he/she will be unable to start the c2d job, because the fees are expired.

Proposal:
 - let's assume that current timestamp is '3June2022 07:00:00'
 - user wants to run a c2d job for 10 minutes
 - user will call initializeCompute, with parameter **duration**='600' (relatime timestamp)
 - user receives instructions how to order dataset & algo and how to pay providerFees.   Provider stores compute env in providerFee.providerData and duration in providerFee.validUntil
 - user generates startOrder transaction(s)
 - once transactions are confirmed, user calls startCompute, with the same **duration** parameter (600 in our case) (or lower if he/she wants)
 - for each asset, provider will perform checks as:
 -  - timestamp = event.timestamp  (does not matter if it's OrderStarted or OrderReused, both have timestamp)
 - -  have a buffer of 60 seconds to manage small block.timestamp possible manipulation
 - - **duration** has to be <= providerFee.validUntil
 - - env has to match providerFee.providerData['environment']
 - - if  timestamp + buffer + **duration** > current_timestamp -> job can start
 - - when calling c2d engine (https://github.com/oceanprotocol/provider/blob/3c03ee209b714ffe536bb995bb3922cd7a448c30/ocean_provider/routes/compute.py#L444), provider will fill payload.maxDuration = **duration**.      (this will replace current payload.validUntil) C2D Engine will handle job cut-off.

Recommandation:
 - make sure that ocean.js/py is sending the transaction that contains providerFee at the end (last transaction from the series) and the other startOrder/reuserOrder transactions which are not containing providerFees are sent before this one

calina-c:
Closes #472


ocean_provider/utils/datatoken.py
```diff
@@ -81,19 +133,23 @@ def verify_order_tx(
...
                 "Mismatch between ordered c2d environment and selected one."
             )
 
-        valid_until = provider_fee_order_log.args.validUntil
-        if (
-            datetime.utcnow().timestamp() >= valid_until
-            and not allow_expired_provider_fees
-        ):
-            raise AssertionError("Ordered c2d time was exceeded, check validUntil.")
+        time_limit = log_datetime + timedelta(
```
alexcos20, line 145: can we add a predefined buffer ?
block.timestamp can be manipulated, so it's not exact

alexcos20:
configured dockerhub, so we can use docker image "oceanprotocol/provider-py:duration" for testing ocean.js/py

ocean_provider/utils/provider_fees.py
```diff
@@ -77,7 +77,7 @@ def get_provider_fees(
         "v": (signed.v + 27) if signed.v <= 1 else signed.v,
         "r": web3.toHex(web3.toBytes(signed.r).rjust(32, b"\0")),
         "s": web3.toHex(web3.toBytes(signed.s).rjust(32, b"\0")),
-        "validUntil": valid_until,
+        "duration": duration,
```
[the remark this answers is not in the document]
calina-c: for provider fee response? Honestly, I disagree. We have seen a huge problem in partially switching from serviceIndex to serviceId and we ended up deciding to only keep the old semantics when dealing with the smart contracts. I think we should use the same reasoning here, by keeping validUntil only in dealing with the smart contracts (and renaming it in smart contracts as soon as possible, when other changes appear)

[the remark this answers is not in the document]
alexcos20: Entire providerFee object has two flows:
 - the lib should pass it through to the contracts
 - if feeAmount>0, then the lib should pay that fee
 By using 'duration' instead of 'validUntil' means that in the lib we have to duplicate the object, change one key (duration -> validUntil) which leads to a lot of confusion.  Also, for devs using something else then ocean.js/ocean.py, this becomes a real hurdle. 
So let's use 'validUntil'

ocean_provider/utils/datatoken.py
alexcos20: add a env called  C2D_BUFFER_DURATION, with a default value of 120 seconds

alexcos20: so time_limit = log_datetime + buffer + timedelta(....

alexcos20 (approved):

alexcos20:
This proposal contains a critical security flow, that would allow users to use c2d with a low amount of fees paid.
Closing this

PR #479, opened 2022-06-08
Adds url redirect checks.

### LINKED ISSUE #478: Redirect checks
#Bug Description
Provider has a way to restrict connections to non-public IP addresses. It is enabled by default with the environment variable ALLOW_NON_PUBLIC_IP=False. Presumably the purpose of this restriction is to protect any services running behind a firewall, and protect provider itself from unauthorised access or denial of service by having provider call itself in a loop.

If a safe URL is passed to provider which results in a 3xx redirect to a non-public IP address, the request will proceed.

#Impact
Could allow an attacker access to provider which has been forbidden. Could allow attacker to probe internal network. Possible information disclosure. Other attacks are possible.

#Risk Breakdown
Difficulty to Exploit: Easy

#Recommendation
Every redirect should be checked in the same way as the original URL.

References
Proof of concept

```
curl -d '{"url":"https://example.com/redirect.php","type":"url","checksum":true}' -H 'Content-Type: application/json' -X POST https://provider/api/services/fileinfo
```

redirect.php:
```
<?php
if($_SERVER['REQUEST_METHOD'] == 'GET') {
        header("HTTP/1.1 301 Moved Permanently");
        header("Location: http://127.0.0.1:8030");
}
exit();
?>
```

calina-c:
Closes #478.

### DIFF
--- a/ocean_provider/utils/url.py
+++ b/ocean_provider/utils/url.py
@@ -20,8 +20,13 @@ REQUEST_TIMEOUT = 3
 CHUNK_SIZE = 8192
 
 
-def get_redirect(url):
+def get_redirect(url, redirect_count=0):
     if not is_url(url):
+        return None
+
+    if redirect_count > 5:
+        logger.info(f"More than 5 redirects for url {url}. Aborting.")
+
         return None
 
     result = requests.head(url, allow_redirects=False)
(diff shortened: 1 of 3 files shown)


ocean_provider/utils/url.py
```diff
@@ -20,8 +20,26 @@
...
+        return None
+
+    result = requests.head(url, allow_redirects=False)
+
+    if result.status_code == 405:
+        # HEAD not allowed, so defaulting to get
+        result = requests.get(url, allow_redirects=False)
+
+    if result.is_redirect:
+        return get_redirect(result.headers["Location"])
```
corrie-sloot, line 34: Redirect may be local, and not a complete URL.

e.g. `Location: /internal-redirect.html`

calina-c: I've added local redirects + tests.

alexcos20 (approved):

PR #480, opened 2022-06-09
Intermediary commit.

calina-c:


calina-c:
Closing since it seems that solving Cloudflare v2 challenges require payment and v1 is deprecated so this PR is not helping right now.

PR #481, opened 2022-06-10
Change unencrypted files structure

calina-c:
Fixes:
  - use new schema (4.1.0)
  - fix compute issues
  - Breaking change:  when calling initialize or initializeCompute, providerFeeAmount is sent as string, and not as integer


ocean_provider/utils/util.py
```diff
@@ -109,29 +110,38 @@ def _generate(_response):
...
 
 def get_service_files_list(
-    service: Service, provider_wallet: LocalAccount, version="4.1.0"
+    service: Service, provider_wallet: LocalAccount, asset: Asset = None
 ) -> list:
-    if version == "4.0.0":
+    if asset is None:
+        version = "4.0.0"
+    else:
+        version = asset.version if asset.version else "4.0.0"
```
calina-c, line 118: or a one-liner: `version = asset.version if asset is not None and asset.version else "4.0.0"`. Although I would rather default to 4.1.0 (CURRENT_VERSION constant)

ocean_provider/utils/util.py
```diff
@@ -109,29 +110,38 @@ def _generate(_response):
...
-        for key in ["datatokenAddress", "type", "files"]:
+        for key in ["datatokenAddress", "nftAddress", "files"]:
             if key not in files_json:
                 raise Exception(f"Key {key} not found in files.")
-
-        if (
-            Web3.toChecksumAddress(files_json["datatokenAddress"])
-            != Web3.toChecksumAddress(service.datatoken_address)
-            or files_json["type"] != service.type
+        if Web3.toChecksumAddress(
```
calina-c, line 132: please leave an empty line before the if, for better readability.

ocean_provider/validation/algo.py
```diff
@@ -344,9 +344,9 @@ def validate(self):
             return False
 
         asset_urls = get_service_files_list(
-            self.service, self.provider_wallet, self.asset.version
+            self.service, self.provider_wallet, self.asset
         )
-
+        logger.error(asset_urls)
```
calina-c, line 349: I don't think this should be error-level, but rather info-level

tests/test_RBAC.py
```diff
@@ -174,7 +176,9 @@ def test_compute_request_payload(
 ):
     ddo_sample1 = copy.deepcopy(ddo_sample1_v4)
     ddo = Asset(ddo_sample1)
-    ddo.services.append(Service.from_json(1, get_compute_service(None, None, "0x0")))
+    ddo.services.append(
+        Service.from_json(1, get_compute_service(None, None, "0x0", "0x0"))
```
calina-c, line 180: please use checksummable addresses here and above, it will save us trouble in the future and I now wonder why these are not checksummed, this should have been detected.

calina-c (commented):
I left a few comments :)

calina-c:
LGTM but I can't approve my own PR.

alexcos20 (approved):

Issue #446, opened 2022-05-13
Enhance Provider Encrypt DDO mechanism to have Fine-grained permission checks


soonhuat:
**Description**
To enhance provider encrypt endpoint to enable market operator to have fine-grained permission check on:

1. Access right check based on wallet
2. DDO values to enforce another layer of restriction based on business use case 

Previously in v3 there is rbac check on data url encryption.
However in v4, it's possible to have 3rd party provider, so data url encryption will be control by 3rd party provider.
In other word, this way of encryption couldn't allow market operator to enforce their market restriction anymore.
With the new publish journey, the encryption fine-grained permission check should happens on DDO encryption.
**assuming market operator will set their provider to perform DDO encryption. 

**Describe the solution**
To have the same rbac checks as v3 but extended version.
Now the same provider endpoint `services/encrypt` will be use for both data url and DDO encryption.
If this stays, the easiest way to differentiate whether is a data urls or DDO would be identify:

1. if body are object instead of array
2. get acocuntId from nft attribute, body.nft.owner

If either of the 2 condition not match, encryption checks should be skipped, assuming that's data urls. 
Provider then send POST request to RBAC-server (along with header if any) with body of the follow structure (which is same as other RBAC checks request structure)
```
{
  eventType: string
  component: string
  authService?: string
  providerAddress?: string
  data?: any
}
```

**RBAC**
RBAC-server will perform:

1. Same credential role access checking as other endpoint
2. Proxy the body or DDO to market operator defined endpoint to perform values checks
3. Return the consolidated truth value of true or false

PR #482, opened 2022-06-11
Issue 1483 upload on publish

calebtuttle:
This is the backend part of the implementation for the spec described in #1483 (and #446).

Changes proposed in this PR:

- Add initializeUpload/ endpoint that returns the fee the Provider charges for upload service.
- Add upload/ endpoint for uploading files. This uploads to Estuary (which pins to IPFS and then establishes deals with Filecoin miners).
- The Provider must have access to an Estuary node. This could be an Estuary node run by the Provider runner or a remote Estuary node. The Estuary team expressed interest in supporting/subsidizing uploads for Ocean Market users.

Additional config variables:
- uploads_path (path to folder where files are temporarily stored before uploaded to external service, e.g., Estuary)
- estuary_api_key
- estuary_url
- upload_fee


AnaLoznianu:
Closing this as out of scope according to the comments in the main issue: https://github.com/oceanprotocol/market/issues/1483

PR #485, opened 2022-06-13
Arweave Integration Milestone 1: Publish using transaction ID

MantisClone:
Towards https://github.com/oceanprotocol/multi-repo-issue/issues/151

Changes proposed in this PR:

- Add new arweave files structure
```
"files": [
  {
    "type": "arweave",
    "transactionId": "a4qJoQZa1poIv5guEzkfgZYSAD0uYm7Vw4zm_tCswVQ",
  }
]
```
- Update `fileinfo` and `download` endpoints to support files stored in Arweave.
- Add `ARWEAVE_GATEWAY` envvar for resolving Arweave transaction ids
- Add `test_download_arweave` to show that downloading an asset stored in Arweave works.
- Add `test_compute_arweave` to show that a compute job that uses a dataset stored in Arweave works.
- Add `test_check_arweave_good` and `test_check_arweave_bad` to show that `fileinfo` endpoint works with arweave transaction ids.

- Fix bug in the `download` endpoint where the Arweave transaction ID or IPFS CID are leaked in the `"Content-Disposition"` response header.
- Discard MismatchedABI warnings that occur when processing event logs - these are expected because a given transaction receipt will likely have many different Events, not just ones that match the Event ABI of interest.
- Add `ARWEAVE_TRANSACTION_ID` constant that points to `branin.arff` stored permanently on Arweave.

> **Note**
> This does not include the ability to upload to Arweave. That is left for a future PR.

Changes that were implemented but then reverted:
- ~~Generalize existing URL files structure (change "url" key to "value") (for details see https://github.com/oceanprotocol/docs/pull/1019#discussion_r902618800)~~


MantisClone:
The CI checks for this PR is failing due to a strange error: https://github.com/oceanprotocol/provider/runs/6887646381?check_suite_focus=true#step:10:8

> Run docker logs ocean_aquarius_1 && docker logs ocean_provider_1 && docker logs ocean_provider2_1 && docker logs ocean_computetodata_1

Does anyone know what might be going wrong?

  docker logs ocean_aquarius_1 && docker logs ocean_provider_1 && docker logs ocean_provider[2](https://github.com/oceanprotocol/provider/runs/6887646381?check_suite_focus=true#step:10:2)_1 && docker logs ocean_computetodata_1
  shell: /usr/bin/bash -e {0}
  env:
    pythonLocation: /opt/hostedtoolcache/Python/[3](https://github.com/oceanprotocol/provider/runs/6887646381?check_suite_focus=true#step:10:3).8.12/x6[4](https://github.com/oceanprotocol/provider/runs/6887646381?check_suite_focus=true#step:10:4)
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.8.12/x[6](https://github.com/oceanprotocol/provider/runs/6887646381?check_suite_focus=true#step:10:6)4/lib
Error: No such container: ocean_aquarius_1
Error: Process completed with exit code 1.

MantisClone:
Lol, I just realized while testing with `ocean.py` that when downloading an arweave file, the filename is the transaction id :facepalm: 

So this PR is not ready. I converted it back to draft. Definitely have to fix it so that the downloaded filename doesn't leak the transaction id.

MantisClone:
This PR is now ready for review @alexcos20 @calina-c

ocean_provider/file_types/definitions.py
```diff
@@ -30,11 +29,7 @@ def check_details(self, with_checksum=False):
         raise NotImplementedError
 
     @abstractmethod
-    def build_download_response(
-        self,
-        request,
-        validate_url=True,
-    ):
+    def build_download_response(self, request):
```
calina-c, line 32: why has the validate_url parameter disappeared from definitions? I understand in IPFS and Arweave we are not using it, but with regular URL files we would like to keep the option of skipping validation if previously validated.

ocean_provider/file_types/file_types.py
```diff
@@ -62,3 +66,41 @@ def get_download_url(self):
...
+        self.transactionId = transactionId
+        self.type = "arweave"
+        self.headers = headers if headers else {}
+        self.userdata = userdata
+        self.method = "get"
+
+    @enforce_types
+    def validate_dict(self) -> Tuple[bool, Any]:
+        if not self.transactionId:
+            return False, "malformed service files, missing required keys."
```
calina-c, line 92: I think you can be more specific here, and say that the transactionId is missing :)

ocean_provider/routes/compute.py
```diff
@@ -116,11 +114,7 @@ def initializeCompute():
     valid_until = int(valid_until)
 
     if not timestamp_ok:
-        return error_response(
-            "The validUntil value is not correct.",
-            400,
-            logger,
-        )
+        return error_response("The validUntil value is not correct.", 400, logger)
```
calina-c, line 117: can you please make sure you use the same black version as we do? We shouldn't have any formatting-only changes.

ocean_provider/routes/consume.py
```diff
@@ -91,12 +84,15 @@ def fileinfo():
...
-    else:
-        files_list = [data]
+    try:
+        if did:
+            asset = get_asset_from_metadatastore(get_metadata_url(), did)
+            service = asset.get_service_by_id(service_id)
+            files_list = get_service_files_list(service, provider_wallet, asset)
+        else:
+            files_list = [data]
+    except Exception as e:
```
calina-c, line 94: I think there is a provider wrapper that returns the error response if any Exception is encountered. Can you please check and remove this if it works as expected?

ocean_provider/utils/datatoken.py
```diff
@@ -3,13 +3,13 @@
...
 from typing import Optional
 
-from jsonsempai import magic  # noqa: F401
 from artifacts import ERC20Template
 from eth_keys import KeyAPI
 from eth_keys.backends import NativeECCBackend
 from eth_typing.encoding import HexStr
 from eth_typing.evm import HexAddress
 from hexbytes import HexBytes
+from jsonsempai import magic  # noqa: F401
```
calina-c, line 12: please take care: you have moved the jsonsempai after the artifacts imports.

ocean_provider/utils/test/test_util.py
```diff
@@ -54,15 +53,8 @@ class Dummy:
...
-    with patch(
-        "ocean_provider.file_types.definitions.is_safe_url",
-        side_effect=[True],
-    ):
-        with patch(
-            "requests.get",
-            side_effect=[mocked_response],
-        ):
-            response = instance.build_download_response(request)
+    with patch("requests.get", side_effect=[mocked_response]):
```
calina-c, line 56: I'm do not agree with mocking the request.get. It is too generic and we risk "testing that our test works". Also, this looks like one of our tests, not an arweave-specific one. It failed? Why would adding the arweave integration warrant changes in the other tests?

ocean_provider/utils/test/test_util.py
```diff
@@ -73,15 +65,8 @@ class Dummy:
...
```
calina-c, line 68: same comment here and in other places throughout this file.

ocean_provider/validation/provider_requests.py
```diff
@@ -266,10 +265,11 @@ def rules(self):
 class FileInfoRequest(CustomJsonRequest):
     def rules(self):
         return {
-            "type": ["required_without:did", "in:ipfs,url"],
+            "type": ["required_without:did", "in:ipfs,url,arweave"],
```
calina-c, line 268: I'm thinking of creating a list called ALLOWED_FILE_TYPES which we can use in instance generation and here as well, to provide a single entrypoint for modifying. Would you like to do that?

I usually guide myself by a rule of "it's ok to do something twice, but the third time I'm automating it"

tests/helpers/compute_helpers.py
```diff
@@ -72,6 +71,29 @@ def build_and_send_ddo_with_compute_service(
...
+        arweave_file_object = {
+            "type": "arweave",
+            "transactionId": "cZ6j5PmPVXCq5Az6YGcGqzffYjx2JnsnlSajaHNr20w",
+        }
+
+        dataset_ddo_w_compute_service = get_registered_asset(
+            publisher_wallet,
+            unencrypted_files_list=[arweave_file_object],
+            custom_services="vanilla_compute",
+            custom_services_args=[
```
calina-c, line 84: this is completely copied. Let's extract a function or assign `standard_compute_args`

tests/test_compute.py
```diff
@@ -297,6 +297,91 @@ def test_compute(client, publisher_wallet, consumer_wallet, free_c2d_env):
...
 
 
+@pytest.mark.integration
+def test_compute_arweave(client, publisher_wallet, consumer_wallet, free_c2d_env):
+    valid_until = get_future_valid_until()
+    ddo, tx_id, alg_ddo, alg_tx_id = build_and_send_ddo_with_compute_service(
+        client,
+        publisher_wallet,
+        consumer_wallet,
+        False,
```
calina-c, line 307: please add the arg name key here for clarity (I assume it's diff_provider or something)

tests/test_compute.py
```diff
@@ -297,6 +297,91 @@ def test_compute(client, publisher_wallet, consumer_wallet, free_c2d_env):
...
+        {
+            "signature": signature,
+            "nonce": nonce,
+            "documentId": ddo.did,
+            "consumerAddress": consumer_wallet.address,
+            "jobId": job_id,
+        }
+    )
+
+    compute_endpoint = BaseURLs.SERVICES_URL + "/compute"
```
calina-c, line 352: do you think we should test the result part? I think starting the compute job is enough and offers enough coverage to your work. If a file is downloadable, then it can be used in compute, so I wouldn't worry for a result.

tests/test_download.py
```diff
@@ -310,3 +310,47 @@ def other_service(_):
                 response.json["error"]
                 == f"Service with index={service.id} is not an access service."
             )
+
+
+def test_download_arweave(client, publisher_wallet, consumer_wallet, web3):
+    unencrypted_files_list = [
+        {
+            "type": "arweave",
+            "transactionId": "cZ6j5PmPVXCq5Az6YGcGqzffYjx2JnsnlSajaHNr20w",
```
calina-c, line 319: I'm not very familiar with arweave, I assume this is a known transaction that you have created. Is it permanent? Can it disappear someday? Can you add some details on that?

tests/test_download.py
```diff
@@ -310,3 +310,47 @@ def other_service(_):
...
+
+    # Consume using index and signature (with nonce)
+    nonce = str(datetime.utcnow().timestamp())
+    _msg = f"{asset.did}{nonce}"
+    payload["signature"] = sign_message(_msg, consumer_wallet)
+    payload["nonce"] = nonce
+    response = client.get(
+        service.service_endpoint + download_endpoint, query_string=payload
+    )
+    assert response.status_code == 200, f"{response.data}"
```
calina-c, line 356: can we assert something about the contents itself?

tests/test_fileinfo.py
```diff
@@ -95,3 +94,38 @@ def test_check_url_bad(client):
...
+@pytest.mark.unit
+def test_check_arweave_bad(client):
+    payload = {"type": "arweave", "transactionId": "invalid"}
+    response = client.post(fileinfo_url, json=payload)
+    result = response.get_json()
+    assert response.status == "200 OK"
+    for file_info in result:
+        assert file_info["valid"] is False
+
+    payload = {"type": "arweave"}  # No transactionId
```
calina-c, line 128: I'm thinking whether there is a possibility we have a transactionId, it is valid, but we have other trouble downloading. Is that possible? What will happen then?

calina-c (changes requested):
Great work, David! And it looks like the recent files refactor is valuable in regards of amount of changes needed to support new files, so I'm happy about that. I've added a few comments and questions.

tests/test_download.py
MantisClone: Sure, I'll add a comment describing the transaction.

Yes, I created this transaction, and assuming the Arweave network operates as intended, it is permanent and will never disappear.  This particular transaction happens to be a blog post that I published using https://mirror.xyz/

Here's the Arweave gateway URL:  https://ogpkhzhzr5kxbkxebt5gazygvm356yr4oythwj4ve2rwq43l3nga.arweave.net/cZ6j5PmPVXCq5Az6YGcGqzffYjx2JnsnlSajaHNr20w

ocean_provider/file_types/definitions.py
MantisClone: I removed the `is_safe_url()` check from `build_download_response` because it was redundant:

> - Remove `is_safe_url` check from `build_download_response`
>   * When calling download, `is_safe_url()` is called earlier by `check_details()` so calling again in `build_download_response` is redunant.
>   * When calling computeResult, `validate_url=False` so `is_safe_url()` check is skipped anyway.

ocean_provider/utils/test/test_util.py
MantisClone: Mocking `requests.get` existed before my changes. I did not add this.

The test failed because I removed the `is_safe_url` check from `build_download_response` and `is_safe_url` nor longer exists at the `ocean_provider.file_types.definitions.is_safe_url` scope.

> - Remove `is_safe_url` check from `build_download_response`
>   * When calling download, `is_safe_url()` is called earlier by `check_details()` so calling again in `build_download_response` is redundant.
>   * When calling computeResult, `validate_url=False` so `is_safe_url()` check is skipped anyway.

@calina-c As I mentioned in my previous comment, I'd rather keep these changes but if it's a deal-breaker, please let me know and I'll revert them.

ocean_provider/utils/test/test_util.py
MantisClone: Please see my response to your comment on line 56 of this file.

ocean_provider/file_types/file_types.py
MantisClone: Done.

ocean_provider/utils/datatoken.py
MantisClone: Done.

ocean_provider/validation/provider_requests.py
MantisClone: Done.

tests/helpers/compute_helpers.py
MantisClone: Done.  I adjusted assets stored in arweave to accept all algorithms (thus eliminating this duplicate code block)

tests/test_compute.py
MantisClone: Done.

tests/test_download.py
MantisClone: I removed all the mentions of the blog post transaction id and instead added `tests/helpers/constants.py` which contains:

```python
"""branin.arff dataset, permanently stored in Arweave"""
ARWEAVE_TRANSACTION_ID = "a4qJoQZa1poIv5guEzkfgZYSAD0uYm7Vw4zm_tCswVQ"
```

tests/test_fileinfo.py
MantisClone: I think I could mock up a test like that.

tests/test_compute.py
MantisClone: I'm okay with removing this result check.

MantisClone: Done.

ocean_provider/routes/compute.py
MantisClone: Done.  I reverted all formatting-only changes.  

I will mention that many of the formatting changes made sense to me.  I don't believe black has been run on all files in this repo.

Additionally, the black version specified in `.pre-commit-config.yaml` does not match the black version specified in `setup.py`

tests/test_download.py
MantisClone: Done.  I asserted the first line of the file.

```python
    assert response.data.decode("utf-8").partition('\n')[0] == "% 1. Title: Branin Function"
```

ocean_provider/routes/consume.py
MantisClone: Done. I have removed the try/except.  Exceptions are caught and handled by the `@app.errorhandler`.

Unfortunately, most of the Exceptions emitted by the provider do not include a `code` attribute so the error response uses 503 `status_code` by default. Updating the Exceptions to use more appropriate status codes is out of scope for this PR.

tests/test_fileinfo.py
MantisClone: I added a test case where the `ARWEAVE_GATEWAY` is invalid and not reachable.  In this case, the response `status` is still `"200 OK"` but the `result[0]["valid"]` is `False`

Any other errors when calling fileinfo or downloading would not be specific to Arweave so I do not believe they'd be in scope for this PR.

MantisClone:
@calina-c I have re-requested a review despite `test_compute` failing. I believe it is failing for reasons outside of this PR. This can be seen in PR https://github.com/oceanprotocol/provider/pull/534. It is "empty", identical to `main`, but `test_compute` is failing in the same way.

ocean_provider/utils/test/test_util.py
calina-c: yes, please keep it for now; we can create an issue to follow-up and discuss on removing is_safe_url, separately.

calina-c:
You can probably fix the failing tests by cherry picking https://github.com/oceanprotocol/provider/pull/533#issuecomment-1235112354

MantisClone:
Thanks, @calina-c You were right, cherry-picking https://github.com/oceanprotocol/provider/commit/6d22a6d5d7b12df5553eb24277bf2ab82ef893e8 fixed the failing test.

I believe this PR is now ready for review.

MantisClone:
@calina-c I have reverted the `is_safe_url` removal and I believe this is ready for your review again.

alexcos20:
Please merge main branch, a new type was added

MantisClone:
Hello @calina-c @alexcos20 I believe this PR is ready for review.

P.S. It seems I am unable to request additional reviewers, perhaps because I am an external contributor.

alexcos20 (approved):
lgtm

PR #493, opened 2022-06-22
Fix InitializeCompute flow

alexcos20:
Changes proposed in this PR:

- make sure we always return correct startOrder tx_id, even if caller sends a reuseOrder tx
- add extra debug logging


calina-c (commented):
Should we keep all the logs?

alexcos20:
> Should we keep all the logs?

I would keep them for a while, until C2d is stable

calina-c (approved):

PR #497, opened 2022-06-23
Bug/fix valid until

### LINKED ISSUE #498: C2D ValidUntil
Our current setup works like:
 - let's assume that current timestamp is '3June2022 07:00:00'
 - user wants to run a c2d job for 10 minutes
 - user will call initializeCompute, with a validUntil = '3June2022 07:10:00'  (absolute timestamp)
 - user receives instructions how to order dataset & algo and how to pay providerFees
 - user generates startOrder/reuseOrder transaction(s)
 - once transactions are confirmed, user calls startCompute, with the same validUntil

But,
What's going to happend if startOrder transactions gets confirmed 30 mins later, due to congested network?
User paid the fees, but he/she will be unable to start the c2d job, because the fees are expired.

Proposal:
 - add 'timestamp' to provider_data.  Equals timestamp of initializeCompute
 - when user calls startCompute , add the difference between tx.timestamp - provider_data['timestamp'] to validUntil  (ie: take into account the time difference between initializeCompute and actual tx confirm time)

alexcos20:
Closes #498

### DIFF
--- a/ocean_provider/utils/datatoken.py
+++ b/ocean_provider/utils/datatoken.py
@@ -95,7 +95,12 @@ def verify_order_tx(
             datetime.utcnow().timestamp() >= valid_until
             and not allow_expired_provider_fees
         ):
-            raise AssertionError("Ordered c2d time was exceeded, check validUntil.")
+            # expired validUntil.  let's add the difference between provider_data["timestamp"] and order timestamp
+            block = web3.eth.get_block(tx_receipt.blockHash, False)
+            if datetime.utcnow().timestamp() >= valid_until + (
+                block.timestamp - provider_data["timestamp"]
+            ):
+                raise AssertionError("Ordered c2d time was exceeded, check validUntil.")
 
     if Web3.toChecksumAddress(
         provider_fee_order_log.args.providerFeeAddress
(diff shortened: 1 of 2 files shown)


ocean_provider/utils/datatoken.py
```diff
@@ -134,13 +129,19 @@ def verify_order_tx(
...
-                "Validity in transaction exceeds current UTC timestamp"
-            )
+        if timestamp_now >= provider_fee_order_log.args.validUntil:
+            # expired validUntil.  let's add the difference between provider_data["timestamp"](time of initializeCompute) and transaction timestamp.
+            # Also add 90 secs as buffer, since block.timestamp can be manipulated
+            block = web3.eth.get_block(tx_receipt.blockHash, False)
+            if provider_initialize_timestamp > 0 and (
+                timestamp_now
+                >= provider_fee_order_log.args.validUntil
+                + (block.timestamp - provider_initialize_timestamp + 90)
```
calina-c, line 142: what does the 90 stand for? is it c2d buffer?

alexcos20: yes. let's keep it hardcoded until we need to change it

calina-c (approved):

PR #505, opened 2022-06-28
Fix fileinfo validation and support headers

### LINKED ISSUE #492: Storage type "URL" should support headers
**Is your feature request related to a problem? Please describe.**
Sometimes, an asset can have files that are protected using some bearer tokens or api keys, which are stored as headers.
A publisher could have this definition for an asset:
```
[
    "type": "url",
    "url": "https://url.com/file1.csv",
    "method": "GET",
    "headers":
      [
       {"Authorization": "Bearer 123"}, 
       {"APIKEY": "124"},
      ]
```
In the current setup, download will fail, because APIKEY header is not sent

Since in https://github.com/oceanprotocol/docs/pull/1029 we added the headers object, Provider should add (if existing) asset defined headers to the request when downloading

### LINKED ISSUE #504: FileInfo Endpoint
Right now, this endpoint supports only 2 scenarios:
 - did  - and provider will decode/decrypt whatever data source is used
 - url

2nd scenario (just url) breaks the current definition of url-storage (see https://docs.oceanprotocol.com/concepts/did-ddo/#files)
We should accept the full object, like:
```
[
    "type": "url",
    "url": "https://url.com/file1.csv",
    "method": "GET",
    "headers":
      [
       {"Authorization": "Bearer 123"}, 
       {"APIKEY": "124"},
]
```

calina-c:
Closes #504. Closes #492.

### DIFF
--- a/ocean_provider/routes/compute.py
+++ b/ocean_provider/routes/compute.py
@@ -532,7 +532,11 @@ def computeResult():
     update_nonce(data.get("consumerAddress"), data.get("nonce"))
 
     response = build_download_response(
-        request, requests_session, result_url, result_url, None, validate_url=False
+        request,
+        requests_session,
+        {"url": result_url, "type": "url"},
+        None,
+        validate_url=False,
     )
     logger.info(f"computeResult response = {response}")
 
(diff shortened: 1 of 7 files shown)


calina-c:
@alexcos20 no restrictions to headers at the current time. Let me know if I should add any.

alexcos20 (approved):
lgtm

PR #516, opened 2022-07-13
Modified encrypt payload for RBAC.

### LINKED ISSUE #446: Enhance Provider Encrypt DDO mechanism to have Fine-grained permission checks

mariacarmina:
Fixes #446  .

Changes proposed in this PR:

- updated RBAC encrypt request.

### DIFF
--- a/ocean_provider/validation/RBAC.py
+++ b/ocean_provider/validation/RBAC.py
@@ -80,37 +80,7 @@ class RBACValidator:
     def get_data(self):
         if "data" not in self.request.keys():
             raise Exception("Data to encrypt is empty.")
-        if not self._is_asset() and not self._is_file_encryption_data():
-            raise Exception("Invalid type of data.")
-
         return self.request["data"]
-
-    def _is_asset(self) -> bool:
-        data = self.request["data"]
-        if not isinstance(data, dict) and not isinstance(data, str):
-            return False
-
-        if isinstance(data, str):
-            try:
-                data = json.loads(data)
-            except:
-                return False
-
-        if isinstance(data, dict):
-            return data.get("version", False)
-
-    def _is_file_encryption_data(self) -> bool:
-        data = self.request["data"]
-        if not isinstance(data, list):
-            return False
-
-        for file in data:
-            if (
-                isinstance(file, dict)
-                and list(file.keys()) == ["nftAddress", "datatokenAddress", "files"]
-                and isinstance(file["files"], dict)
-            ):
-                return True
 
     def build_payload(self):
         provider_access = (
--- a/ocean_provider/validation/RBAC.py
+++ b/ocean_provider/validation/RBAC.py
@@ -80,36 +80,37 @@ class RBACValidator:
     def get_data(self):
         if "data" not in self.request.keys():
             raise Exception("Data to encrypt is empty.")
-        if (
-            not self._check_if_asset()
-            and not self._check_if_respects_file_encryption_schema()
-        ):
+        if not self._is_asset() and not self._is_file_encryption_data():
             raise Exception("Invalid type of data.")
 
         return self.request["data"]
 
-    def _check_if_asset(self):
+    def _is_asset(self) -> bool:
         data = self.request["data"]
-        if isinstance(data, dict):
-            if data.get("version") is not None:
-                return True
-        elif isinstance(data, str):
-            data_dict = json.loads(data)
-            if data_dict.get("version") is not None:
-                return True
-        return False
+        if not isinstance(data, dict) and not isinstance(data, str):
+            return False
 
-    def _check_if_respects_file_encryption_schema(self):
+        if isinstance(data, str):
+            try:
+                data = json.loads(data)
+            except:
+                return False
+
+        if isinstance(data, dict):
+            return data.get("version", False)
+
+    def _is_file_encryption_data(self) -> bool:
         data = self.request["data"]
-        if isinstance(data, list):
-            for file in data:
-                if (
-                    isinstance(file, dict)
-                    and list(file.keys()) == ["nftAddress", "datatokenAddress", "files"]
-                    and isinstance(file["files"], dict)
-                ):
-                    return True
-        return False
+        if not isinstance(data, list):
+            return False
+
+        for file in data:
+            if (
+                isinstance(file, dict)
+                and list(file.keys()) == ["nftAddress", "datatokenAddress", "files"]
+                and isinstance(file["files"], dict)
+            ):
+                return True
 
     def build_payload(self):
         provider_access = (
--- a/ocean_provider/validation/RBAC.py
+++ b/ocean_provider/validation/RBAC.py
@@ -10,7 +10,6 @@ import requests
 
 from ocean_provider.exceptions import RequestNotFound
 from ocean_provider.utils.accounts import sign_message
-from ocean_provider.utils.asset import Asset
 from ocean_provider.utils.basics import get_provider_wallet
 
 
@@ -81,12 +80,36 @@ class RBACValidator:
     def get_data(self):
         if "data" not in self.request.keys():
             raise Exception("Data to encrypt is empty.")
-        if not isinstance(self.request["data"], list) and not isinstance(
-            self.request["data"], Asset
+        if (
+            not self._check_if_asset()
+            and not self._check_if_respects_file_encryption_schema()
         ):
             raise Exception("Invalid type of data.")
 
         return self.request["data"]
+
+    def _check_if_asset(self):
+        data = self.request["data"]
+        if isinstance(data, dict):
+            if data.get("version") is not None:
+                return True
+        elif isinstance(data, str):
+            data_dict = json.loads(data)
+            if data_dict.get("version") is not None:
+                return True
+        return False
+
+    def _check_if_respects_file_encryption_schema(self):
+        data = self.request["data"]
+        if isinstance(data, list):
+            for file in data:
+                if (
+                    isinstance(file, dict)
+                    and list(file.keys()) == ["nftAddress", "datatokenAddress", "files"]
+                    and isinstance(file["files"], dict)
+                ):
+                    return True
+        return False
 
     def build_payload(self):
         provider_access = (
--- a/ocean_provider/validation/RBAC.py
+++ b/ocean_provider/validation/RBAC.py
@@ -81,7 +81,6 @@ class RBACValidator:
     def get_data(self):
         if "data" not in self.request.keys():
             raise Exception("Data to encrypt is empty.")
-        # import pdb;pdb.set_trace()
         if not isinstance(self.request["data"], list) and not isinstance(
             self.request["data"], Asset
         ):
--- a/ocean_provider/validation/RBAC.py
+++ b/ocean_provider/validation/RBAC.py
@@ -10,6 +10,7 @@ import requests
 
 from ocean_provider.exceptions import RequestNotFound
 from ocean_provider.utils.accounts import sign_message
+from ocean_provider.utils.asset import Asset
 from ocean_provider.utils.basics import get_provider_wallet
 
 
@@ -77,6 +78,17 @@ class RBACValidator:
             for additional_input in additional_inputs
         ]
 
+    def get_data(self):
+        if "data" not in self.request.keys():
+            raise Exception("Data to encrypt is empty.")
+        # import pdb;pdb.set_trace()
+        if not isinstance(self.request["data"], list) and not isinstance(
+            self.request["data"], Asset
+        ):
+            raise Exception("Invalid type of data.")
+
+        return self.request["data"]
+
     def build_payload(self):
         provider_access = (
             "private" if os.getenv("PRIVATE_PROVIDER", False) else "public"
@@ -97,7 +109,7 @@ class RBACValidator:
         message = "encryptUrl" + json.dumps(self.credentials)
         signature = sign_message(message, get_provider_wallet())
 
-        return {"signature": signature, "data": self.request["data"]}
+        return {"signature": signature, "data": self.get_data()}
 
     def build_initialize_payload(self):
         message = "initialize" + json.dumps(self.credentials)
--- a/ocean_provider/validation/RBAC.py
+++ b/ocean_provider/validation/RBAC.py
@@ -97,7 +97,7 @@ class RBACValidator:
         message = "encryptUrl" + json.dumps(self.credentials)
         signature = sign_message(message, get_provider_wallet())
 
-        return {"signature": signature}
+        return {"signature": signature, "data": self.request["data"]}
 
     def build_initialize_payload(self):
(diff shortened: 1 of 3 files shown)


ocean_provider/validation/RBAC.py
```diff
@@ -77,6 +78,17 @@ def get_additional_dids(self):
             for additional_input in additional_inputs
         ]
 
+    def get_data(self):
+        if "data" not in self.request.keys():
+            raise Exception("Data to encrypt is empty.")
+        # import pdb;pdb.set_trace()
+        if not isinstance(self.request["data"], list) and not isinstance(
```
soonhuat, line 85: while request /encrypt hit, both asset files encryption and ddo encryption will send to RBAC and let RBAC decide what to do next?

soonhuat:
@mariacarmina should https://github.com/oceanprotocol/provider/blob/657b177ed1d30899a12c15dca4be3d3027541550/ocean_provider/routes/encrypt.py#L22
 have @validate(EncryptRequest) decorator?

ocean_provider/validation/RBAC.py
```diff
@@ -77,6 +78,16 @@ def get_additional_dids(self):
             for additional_input in additional_inputs
         ]
 
+    def get_data(self):
+        if "data" not in self.request.keys():
+            raise Exception("Data to encrypt is empty.")
+        if not isinstance(self.request["data"], list) and not isinstance(
```
calina-c, line 84: please add an empty line before the second `if`, for readability. How come these can be only list and Asset? Is it possible this is a dict or a string json that needs conversion to Asset?

ocean_provider/validation/RBAC.py
```diff
@@ -77,6 +78,16 @@ def get_additional_dids(self):
...
+            self.request["data"], Asset
+        ):
+            raise Exception("Invalid type of data.")
```
calina-c, line 87: I'm not big on raising an Exception here. Is it possible to not have data? Previously we had this option, so it should be still supported? i.e. if there is no data, do not send it?

I would also recommend changing the function name to get_request_data, which is clearer.

tests/test_RBAC.py
```diff
@@ -45,11 +46,12 @@ def test_encrypt_request_payload(consumer_wallet, publisher_wallet, monkeypatch)
         "compression": "zip",
     }
     req = {
-        "document": json.dumps(document),
+        "data": [document],
```
calina-c, line 49: we shhould definitely still support the version with json.dumps, this should be fully compatible.

calina-c (commented):
I think there is still some confusion about what needs to be done here. Can you guys please share some discussions/conclusions? @soonhuat @mariacarmina because I might be misguided in what I'm reviewing exactly.

soonhuat:
> I think there is still some confusion about what needs to be done here. Can you guys please share some discussions/conclusions? @soonhuat @mariacarmina because I might be misguided in what I'm reviewing exactly.

@calina-c to extend RBAC to validate metadata data structure or values before encryption of metadata.
I assume this RBAC extended checks should only happens during metadata encryption, but not during asset url encryption, since now encryption endpoint are shared for both type of encryption (metadata and asset url)

mariacarmina:
I have understood so far that provider should be a pass-through for RBAC server. Is it necessary to validate the type for `data` before sending to RBAC if RBAC will do the distinction between asset URL (files list) and the actual asset object. In the first place, I have done the checks to see what kind of data I send to RBAC, but for asset encryption it won't work, because I have a specific `Asset` object. For example, it won't work from ocean.js side, because the objects won't be compatible with the python object `Asset`. In my perspective, if we want to keep the checks we should check if the asset object is actually a JSON or a stringified JSON. What are your thoughts here @soonhuat @calina-c?

soonhuat:
> won't be compatible with the python object `Asset`

@mariacarmina if Asset object wouldn't compatible is request were from oceanJs or market, how about check against if the `data` is _files (asset URL)_ instead, then skip RBAC if is _files_ structure.

mariacarmina:
> > won't be compatible with the python object `Asset`
> 
> @mariacarmina if Asset object wouldn't compatible is request were from oceanJs or market, how about check against if the `data` is _files (asset URL)_ instead, then skip RBAC if is _files_ structure.

Ok I have understood, so RBAC should accept a list of URLs, not of file structure object, but how about the Asset object case? If provider receives an asset for encryption, in which form should be exactly? A disctionary or a stringified JSON?

soonhuat:
> RBAC should accept a list of URLs

`so RBAC should accept a list of URLs, not of file structure object` , I guess no, because provider will skip RBAC if is `asset files ` structure right?

As for scenario when provider encrypt request input are DDO, then RBAC will receive as it is, assuming RBAC will then (proxy to another validation API) or (market operator will fork the RBAC handle the validation themselves), finally return true/false.

mariacarmina:
> As for scenario when provider encrypt request input are Metadata/Asset, then RBAC will receive as it is, assuming RBAC will then (proxy to another validation API) or (market operator will fork the RBAC handle the validation themselves), finally return true/false.

What means `then RBAC will receive as it is`? The `Asset` object from provider is not compatible with Asset object from ocean.js for example, is the same situation as it is with the file structure object, that's why I assume that provider should check first if the asset object is serialized or is a dictionary.

soonhuat:
> object from provider is not compatible with Asset object from ocean.js for example, is the same situation as it is with the file structure object

@mariacarmina yup, agree (provider should check first if the asset object is serialized or is a dictionary), looks like there is similar check around 71-72
https://github.com/oceanprotocol/provider/blob/e89d6a09c5dd077fba6b8377c4a231aa974e2ea0/ocean_provider/utils/util.py#L72
maybe this is the checking u mean?

after the check, if is NOT data files encrypt, then just like your PR : POST the request body to RBAC (the `then RBAC will receive as it is` I mean previously)

mariacarmina:
> 

Yes, I have added the proper checks as you mentioned in your description. I will do the changes likewise and provide additional checks for `Asset` if it's serialized already or a dictionary. Regarding the check for files list, I need to be more specific for it because it should only contains list of URLs, not just a simple list.

ocean_provider/validation/RBAC.py
```diff
@@ -77,6 +77,40 @@ def get_additional_dids(self):
...
+            and not self._check_if_respects_file_encryption_schema()
+        ):
+            raise Exception("Invalid type of data.")
+
+        return self.request["data"]
+
+    def _check_if_asset(self):
+        data = self.request["data"]
+        if isinstance(data, dict):
+            if data.get("version") is not None:
```
calina-c, line 94: if it is a dict you can directly `return data.get("version", False)`

ocean_provider/validation/RBAC.py
```diff
@@ -77,6 +77,40 @@ def get_additional_dids(self):
...
+        if (
+            not self._check_if_asset()
+            and not self._check_if_respects_file_encryption_schema()
+        ):
+            raise Exception("Invalid type of data.")
+
+        return self.request["data"]
+
+    def _check_if_asset(self):
+        data = self.request["data"]
```
calina-c, line 92: I would add a preliminary check: if it is not dict, nor string, return False directly, to avoid if imbrication later

ocean_provider/validation/RBAC.py
```diff
@@ -77,6 +77,40 @@ def get_additional_dids(self):
...
+        if isinstance(data, dict):
+            if data.get("version") is not None:
+                return True
+        elif isinstance(data, str):
+            data_dict = json.loads(data)
```
calina-c, line 97: you need to add exception handling for the json.loads, e.g. what if the string is "bla bla"? That can not be loaded into a json dict. I would also change the order like this:
- perform json loading if string. if the json conversion fails, we return False
- the resulting dict (coming from the string loading or directly on the payload) needs the same version check and that's it. No code duplication.

ocean_provider/validation/RBAC.py
```diff
@@ -77,6 +77,40 @@ def get_additional_dids(self):
...
+            if data_dict.get("version") is not None:
+                return True
+        return False
+
+    def _check_if_respects_file_encryption_schema(self):
+        data = self.request["data"]
+        if isinstance(data, list):
```
calina-c, line 104: can it be a stringified list?
I don't like the imbrication here, it is not clear to me at first glance what this does. This means you need to improve its readability.

tests/test_RBAC.py
```diff
@@ -58,6 +56,44 @@ def test_encrypt_request_payload(consumer_wallet, publisher_wallet, monkeypatch)
...
+    ddo = copy.deepcopy(json_dict)
+    req["data"] = ddo
+
+    validator = RBACValidator(request_name="EncryptRequest", request=req)
+    payload = validator.build_payload()
+    assert payload["data"] == ddo
+
+
+@pytest.mark.unit
+def test_wrong_encrypt_request_payload(consumer_wallet, publisher_wallet, monkeypatch):
```
calina-c, line 70: please add tests for the individual functions as well, the test coverage is very poor on your added code. How do I know this? The imbrications many and hard to read. Poorly readable means poorly testable, so that's why you couldn't add tests.

calina-c (changes requested):
I'm sure this works, but you can do much better.

ocean_provider/validation/RBAC.py
mariacarmina: What if we can verify in the strigified json if the `version: "4.1.0"` exists? Without converting into a dictionary.

calina-c: It should cover all the cases, but if it is not a dictionary then it is invalid anyway, so I would check both.

ocean_provider/validation/RBAC.py
mariacarmina: @soonhuat is it necessary to check a stringified list of files in provider?

soonhuat: I think don't need to the stringified list check
TODAY if provider interface from market or oceanJs, it won't be a stringify list, is always either stringify of DDO object or asset file object as below structure (schema 4.1.0):
{
    nftAddress,
    datatokenAddress,
    files: [
      {
        type: 'url',
        index: 0,
        url: "abcdefg",
        method: 'GET'
      }
    ]
}
based on calina suggestion, maybe rename it to _is_file_encryption_data would able to self explaining itself

soonhuat: data is not a list, is an object
and probably line 107 and 108 is enough already, since that's exactly is asset file encryption data structure

ocean_provider/validation/RBAC.py
```diff
@@ -77,6 +77,41 @@ def get_additional_dids(self):
...
+            return False
+
+        if isinstance(data, str):
+            try:
+                data = json.loads(data)
+            except:
+                return False
+
+        if isinstance(data, dict):
+            return data.get("version", False)
```
calina-c, line 100: I would convert this to a bool.

ocean_provider/validation/RBAC.py
```diff
@@ -77,6 +77,41 @@ def get_additional_dids(self):
...
+        if not isinstance(data, list):
+            return False
+
+        for file in data:
+            if (
+                isinstance(file, dict)
+                and list(file.keys()) == ["nftAddress", "datatokenAddress", "files"]
+                and isinstance(file["files"], dict)
+            ):
+                return True
```
calina-c, line 113: shouldn't this be higher-level? Is it a file structure if just only one of the elements in data has the correct structure?

can you also look into the list comparison possible issues? What if I add the nftAddress, datatokenAddress and files in a different order? Will comparing with `==` work?

can files also be stringified?

I think we should only handle the data encryption separately, and leave the rest as is, without restrictions. As far as I know, anything should be encryptable.

calina-c: And actually, people can create their own private aquariuses with different schemas, so we shouldn't do any validation on this structure.

tests/test_RBAC.py
calina-c: this is still a thing, I'll add a generic code review comment explaining what I recommend.

tests/test_RBAC.py
```diff
@@ -58,6 +56,58 @@ def test_encrypt_request_payload(consumer_wallet, publisher_wallet, monkeypatch)
...
+    monkeypatch.setenv("PRIVATE_PROVIDER", "1")
+    req = {
+        "publisherAddress": publisher_wallet.address,
+    }
+    validator = RBACValidator(request_name="EncryptRequest", request=req)
+    with pytest.raises(Exception) as err:
+        validator.build_payload()
+    assert err.value.args[0] == "Data to encrypt is empty."
```
calina-c, line 78: you can simplify testing using `match` as an argument of `raises` instead of `err.value.args`

calina-c (changes requested):
Here is my 2c on this. Since this has gone through so many iterations, I would like to scrap it altogether and copy just the data-related part in a new PR. I don't think we should impose any restrictions on the payload. No validation, nothing extra. All we need to do is detect if there is an asset structure using the `_is_asset` function and send it to rbac. That's it. And it will be more efficient to copy that over into a new PR, for easier tracking of changes.

After copying the `_is_asset` function, add tests for that function particularly. Then a couple of extra test in the rbac file: when data should be sent and when not.

soonhuat (approved):

soonhuat:
Looks good, if https://github.com/oceanprotocol/provider/blob/f6bc2e2a1ed19361d5bd6a5ea20ca5f766cc3075/ocean_provider/routes/encrypt.py#L21
doesn't need validate decorator like `@validate(EncryptRequest) `

PR #517, opened 2022-07-15
Adds paid environment compute test.

### LINKED ISSUE #464: Add tests to use the new C2D environment(the one that has costs)

calina-c:
Closes #464.

### DIFF
--- a/tests/test_compute.py
+++ b/tests/test_compute.py
@@ -543,7 +543,9 @@ def test_compute_environments(client):
 
 
 @pytest.mark.integration
-def test_compute_paid_env(client, publisher_wallet, consumer_wallet, paid_c2d_env, web3):
+def test_compute_paid_env(
+    client, publisher_wallet, consumer_wallet, paid_c2d_env, web3
+):
     valid_until = get_future_valid_until()
     deployer_wallet = Account.from_key(os.getenv("FACTORY_DEPLOYER_PRIVATE_KEY"))
     fee_token = get_datatoken_contract(web3, get_ocean_token_address(web3))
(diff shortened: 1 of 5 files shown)


tests/test_compute.py
```diff
@@ -535,3 +540,53 @@ def test_compute_environments(client):
...
+        "signature": signature,
+        "nonce": nonce,
+        "consumerAddress": consumer_wallet.address,
+        "environment": paid_c2d_env["id"],
+        "signature": signature,
+    }
+
+    # Start compute with valid signature
+    response = post_to_compute(client, payload)
+    assert response.status == "200 OK", f"start compute job failed: {response.data}"
```
mariacarmina, line 588: Can we have a test where this compute env is not paid? What error message is thrown in that case?

calina-c: Sure, I'll do it :)

calina-c: @mariacarmina The compute tests are generally integration tests which combine a startOrder with a compute start job. In this case, the startOrder will fail from contracts, which is to be expected. Our initialize endpoint only provides the necessary data for the user to make the startOrder call (through .py, .js or contracts). This means that an unpaid compute test wouldn't be a test of initialize, nor computeStart, but of internal functions from contracts/py/js. I can add it if you are adamant, but instead I added the following:

- a correction to showcase that initializeCompute gives amount=0 for unpaid envs
- a test to show that initializeCompute gives an expected amount for paid envs
- the end-to-end compute test was unmodified and uses the internal startOrder function, as previously

Also, please note that in tests_validation we have a test called `test_fee_amount_not_paid` which fills in the blanks between these two. Basically, it is now tested for paid envs that initializeCompute behaves ok and a combination of correct startOrder + compute job start also works ok. Testing the startOrder functionality itself, I believe it is not in our scope, but let me know what you think.

PR #525, opened 2022-08-10
Validation errors

### LINKED ISSUE #523: Add a status code on C2D errors

calina-c:
Closes #523

### DIFF
--- a/ocean_provider/validation/algo.py
+++ b/ocean_provider/validation/algo.py
@@ -24,7 +24,7 @@ from ocean_provider.utils.util import (
     get_service_files_list,
     msg_hash,
 )
-from ocean_provider.validation.images import validate_docker
+from ocean_provider.validation.images import validate_container
 
 logger = logging.getLogger(__name__)
 
@@ -344,7 +344,7 @@ def validate_formatted_algorithm_dict(algorithm_dict, algorithm_did):
         )  # noqa
 
     container = algorithm_dict.get("container", {})
-    valid, message = validate_docker(container)
+    valid, message = validate_container(container)
     if not valid:
         return False, "container", message
 
--- a/ocean_provider/validation/algo.py
+++ b/ocean_provider/validation/algo.py
@@ -71,7 +71,8 @@ class WorkflowValidator:
             additional_inputs = []
 
         if not isinstance(additional_inputs, list):
-            self.error = "additional_input.invalid"
+            self.resource = "additional_input"
+            self.message = "invalid"
             return False
 
         all_data = [main_input] + additional_inputs
@@ -100,8 +101,8 @@ class WorkflowValidator:
 
             status = input_item_validator.validate()
             if not status:
-                prefix = f"datasets[{index}]." if index else "dataset."
-                self.error = prefix + input_item_validator.error
+                self.resource = input_item_validator.resource
+                self.message = input_item_validator.message
                 return False
 
             self.validated_inputs.append(input_item_validator.validated_inputs)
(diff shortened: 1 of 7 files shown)


calina-c:
I'm now thinking about another possible improvement: the error returned can be a key => value mapping:
- the key would be the field it arised on, nested with dots
- the value would be the error itself

@bogdanfazakas @alexcos20 would you like that? I'd like this intermediary version reviewed, tested and merged though, since it contains structural changes as well.

ocean_provider/validation/algo.py
```diff
@@ -253,12 +246,79 @@ def _build_and_validate_algo(self, algo_data):
...
+
+    return algo_files_checksum, algo_container_checksum
+
 
 def validate_formatted_algorithm_dict(algorithm_dict, algorithm_did):
     if algorithm_did and not (
         algorithm_dict.get("url") or algorithm_dict.get("remote")
     ):
-        return False, f"cannot get url for the algorithmDid {algorithm_did}"
+        return False, "algorithm.did_not_found"
```
alexcos20, line 321: algorithm.did_not_found  means that there is no such did

alexcos20 (approved):

alexcos20:
I support key => value errors descriptions, it makes it more readable by devs

ocean_provider/validation/algo.py
```diff
@@ -276,11 +336,11 @@ def validate_formatted_algorithm_dict(algorithm_dict, algorithm_did):
         if not container.get(key):
             return (
                 False,
-                "algorithm `container` must specify values for all of entrypoint, image and checksum.",
+                "algorithm.container.missing_entrypoint_image_checksum",
             )
 
     if not container["checksum"].startswith("sha256:"):
-        return False, "container checksum must start with sha256:"
+        return False, "algorithm.container.checksum_prefix"
```
bogdanfazakas, line 343: In my opinion from the client side perspective it would make more sense and would be easier to deal with the error on the market for example if we have the error structured in a more consistent way like `resource.action`. like we have here for eg `algorithm.file_unavailable` where the left hand side is the resource and the right one is the action. And would be nice to split the resources as we split the action

So for this case i see the error as following:
```suggestion
        return False, "algorithm_container.checksum_prefix"
```

ocean_provider/validation/algo.py
```diff
@@ -359,18 +419,18 @@ def validate(self):
...
-            self.error = "Service for main asset must be compute."
+            self.error = "main_service_compute"
             return False
 
         asset_urls = get_service_files_list(
             self.service, self.provider_wallet, self.asset
         )
         if self.service.type == "compute" and not asset_urls:
-            self.error = "Services in input with compute type must be in the same provider you are calling."
+            self.error = "compute_services_not_in_same_provider"
```
bogdanfazakas, line 433: Based on the previous comment, not really sure what would be the resource and the action here. Maybe `compute_services.not_in_same_provider`

bogdanfazakas:
> I support key => value errors descriptions, it makes it more readable by devs

I agree with Alex here, so besides the improvements from this PR would be nice to have` key => value` errors descriptions and use the new strings as `values` and some error codes as `keys`

calina-c:
@bogdanfazakas is it ok with you if we merge this, since it contains structural changes? Then add in the key value logic?

bogdanfazakas:
> @bogdanfazakas is it ok with you if we merge this, since it contains structural changes? Then add in the key value logic?

The current state does not make our life easier on the client side, what about the previous suggestions from the PR ?, I'm fine with adding the key value logic later but, let's discuss the other stuff before merging.

alexcos20 (approved):
lgtm, small issues that we can handle later
@bogdanfazakas - any last minute comments?

bogdanfazakas:
> lgtm, small issues that we can handle later @bogdanfazakas - any last minute comments?

lg, I think we can handle the changes in the market, does this really close #523 also ?

calina-c:
> lg, I think we can handle the changes in the market, does this really close #523 also ?

yep, all error messages are now key-value pairs with status messages :)

PR #540, opened 2022-09-16
Feature/smartcontract data service

alexcos20:
Allow smartcontract view functions or public variables to be used as data source for any asset

# Simple call (no user input required) 
This will call swapOceanFee function on the router contract and use the output as source for a dataset
Example:

```
abi = {
        "inputs": [],
        "name": "swapOceanFee",
        "outputs": [{"internalType": "uint256", "name": "", "type": "uint256"}],
        "stateMutability": "view",
        "type": "function",
    }
    unencrypted_files_list = [
        {"type": "smartcontract", "address": "0x8149276f275EEFAc110D74AFE8AFECEaeC7d1593", "abi": abi}
    ]
```

# Parametrized call
(requires user_parameters that matches abi input elements)

This will call getOPCFee function on the router contract and use the output as source for a dataset.
getOPCFee function requires one parameter, called baseToken, so we need to define that in list of custom_parameters, so consumer specifies a value
```
abi = {
        "inputs": [{"internalType": "address", "name": "baseToken", "type": "address"}],
        "name": "getOPCFee",
        "outputs": [{"internalType": "uint256", "name": "", "type": "uint256"}],
        "stateMutability": "view",
        "type": "function",
    }
    unencrypted_files_list = [
        {"type": "smartcontract", "address": "0x8149276f275EEFAc110D74AFE8AFECEaeC7d1593", "abi": abi}
    ]
    asset = get_registered_asset(
        publisher_wallet,
        unencrypted_files_list=unencrypted_files_list,
        custom_userdata=[
            {
                "name": "baseToken",
                "type": "text",
                "label": "baseToken",
                "required": True,
                "description": "baseToken to check for fee",
            }
        ],
    )
```

FYI:
 - only stateMutability = view is supported (both functions and public smartcontract variables).  Provider cannot call functions that needs a transaction to be generated
 - only one element of abi is required  (function name is extracted from there), so you cannot paste the entire smartcontract abi
 - if there are inputs variables required by the function, then those variables must be defined as custom_parameters

### DIFF
--- a/ocean_provider/file_types/types/smartcontract.py
+++ b/ocean_provider/file_types/types/smartcontract.py
@@ -67,17 +67,14 @@ class SmartContractCall(FilesType):
             address=web3.toChecksumAddress(self.address), abi=[self.abi]
         )
         function = contract.functions[self.abi.get("name")]
-        args = dict() if len(self.abi.get("inputs")) > 0 else None
+        args = dict()
         for input in self.abi.get("inputs"):
             args[input.get("name")] = self.userdata.get(input.get("name"))
             if input.get("type") == "address":
                 args[input.get("name")] = web3.toChecksumAddress(
                     args[input.get("name")]
                 )
-        if args:
-            result = function(**args).call()
-        else:
-            result = function().call()
+        result = function(**args).call()
         if isinstance(result, object):
             return json.dumps(result), "application/json"
         return result, "application/text"
--- a/ocean_provider/file_types/types/smartcontract.py
+++ b/ocean_provider/file_types/types/smartcontract.py
@@ -90,10 +90,7 @@ class SmartContractCall(FilesType):
     def check_details(self, with_checksum=False):
         try:
             result, type = self.fetch_smartcontract_call()
-            details = {
-                "contentLength": len(result) or "",
-                "contentType": type
-            }
+            details = {"contentLength": len(result) or "", "contentType": type}
             return True, details
         except Exception as e:
             return False, {}
--- a/ocean_provider/file_types/types/smartcontract.py
+++ b/ocean_provider/file_types/types/smartcontract.py
@@ -67,11 +67,7 @@ class SmartContractCall(FilesType):
             address=web3.toChecksumAddress(self.address), abi=[self.abi]
         )
         function = contract.functions[self.abi.get("name")]
-        if len(self.abi.get("inputs")) > 0:
-            args = dict()
-        else:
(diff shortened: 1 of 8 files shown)


ocean_provider/file_types/types/smartcontract.py
```diff
@@ -0,0 +1,110 @@
...
+    def get_filename(self) -> str:
+        return uuid4().hex
+
+    def fetch_smartcontract_call(self):
+        web3 = get_web3()
+        contract = web3.eth.contract(
+            address=web3.toChecksumAddress(self.address), abi=[self.abi]
+        )
+        function = contract.functions[self.abi.get("name")]
+        if len(self.abi.get("inputs")) > 0:
```
calina-c, line 70: i suggest the one-liner here: ```args = dict() if len(self.abi.get("inputs")) > 0 else None```

ocean_provider/file_types/types/smartcontract.py
```diff
@@ -0,0 +1,110 @@
...
+            args = None
+
+        for input in self.abi.get("inputs"):
+            args[input.get("name")] = self.userdata.get(input.get("name"))
+            if input.get("type") == "address":
+                args[input.get("name")] = web3.toChecksumAddress(
+                    args[input.get("name")]
+                )
+        if args:
+            result = function(**args).call()
```
calina-c, line 82: Can you try whether an empty dict works with the `**` syntax? I think it should, and then you can remove the dict/None initialization in lines 70-73.

ocean_provider/file_types/types/smartcontract.py
```diff
@@ -0,0 +1,110 @@
...
+        else:
+            result = function().call()
+        if isinstance(result, object):
+            return json.dumps(result), "application/json"
+        else:
```
calina-c, line 87: no need for the else if the only instruction inside the if is a return :)

ocean_provider/file_types/types/smartcontract.py
```diff
@@ -0,0 +1,110 @@
...
+            return result, "application/text"
+
+    def check_details(self, with_checksum=False):
```
calina-c, line 90: you can also assign the result of the fetch smart contract call to an attribute and only retrieve it in build_download_response if it does not exist.

calina-c (changes requested):
Some changes requested :)

calina-c (approved):

PR #557, opened 2022-10-28
Documentation for possible provider errors

### LINKED ISSUE #548: Document provider errors
- [ ] create a README or a file to document the errors and the status codes that the provider returns
- [ ] add a link in the docs https://github.com/oceanprotocol/docs/issues/1143
- [ ] Provider related issue: https://github.com/oceanprotocol/provider/issues/523

mariacarmina:
Works on #548  .

Changes proposed in this PR:

- create a separate README which contains explained status codes with the description added.


ocean_provider/routes/README.md
```diff
@@ -0,0 +1,166 @@
...
+```
+or
+
+```python
+{
+    "algorithm.serviceId": "not_found"
+}
+```
+
+**Reason** The `serviceId` key is missing from the algorithm's DDO.
```
calina-c, line 82: not really the same thing: "missing" means the ket is missing from the user input, so there's no way of retrieving the service. "not_found" means the key is present, but doesn't exist. i.e. there is service Id "A" and "B" on the DDO, but the user requested service "C".

ocean_provider/routes/README.md
```diff
@@ -0,0 +1,166 @@
...
+
+```python
+{
+    "algorithm.documentId": "did_not_found"
+}
+```
+
+**Reason** The asset `DID` could not be retrieved from the metadata store.
```
calina-c, line 90: please specify here that this is the algorithm's DID and the "algorithm asset" does not exist, to avoid unclarity regarding which asset is unretrievable

ocean_provider/routes/README.md
```diff
@@ -0,0 +1,166 @@
...
+```
+**Reason** Data NFT address is not present in the asset object.
+
+```python
+{
+    "error": "Asset is not consumable."
+}
+```
+
+**Reason** Metadata status is not in the range of valid status codes for
```
calina-c, line 105: please be more specific here

ocean_provider/routes/README.md
```diff
@@ -0,0 +1,166 @@
...
+**Reason** If the main service is not `compute` for this asset when calling `initialize`
+endpoint.
+
+```python
+{
+    "algorithm.serviceId": "compute_services_not_in_same_provider"
+}
+```
+
+**Reason** Files attached to the compute service are not decrypted by the correct provider.
```
calina-c, line 138: please be more specific here, this is only in the case where both asset and algo are requested by their compute service

ocean_provider/routes/README.md
```diff
@@ -0,0 +1,166 @@
...
+}
+```
+**Reason** `validUntil` value is most probably expired.
+
+```python
+{
+    "error": "Compute environment does not exist."
+}
+```
+**Reason** If you are trying to initialize a compute service, the compute values need to be provided.
```
calina-c, line 152: This is not it. Here, the compute environment provided by the user does not exist, it is not served by our c2d. The user can use get_c2d_environments to check the list of available compute envs.

ocean_provider/routes/README.md
```diff
@@ -0,0 +1,166 @@
...
+```python
+{
+    "error": "DID is not a valid algorithm."
+}
+```
+**Reason** Either the DID is incorrect typed, either the algorithm timeout expired.
```
calina-c, line 158: please specify we are talking about the algorithm asset

calina-c (changes requested):
left a few comments

ocean_provider/routes/README.md
calina-c: also, it's "incorrectly" typed, not "incorrect" typed

calina-c (approved):

PR #576, opened 2022-11-24
Remove Config object.

calina-c:


alexcos20:
looks good, but I think you have some glitches when passing chainId to compute service, and that's why tests are failing

calina-c:
> looks good, but I think you have some glitches when passing chainId to compute service, and that's why tests are failing

The tests are failing because the multichain branch is not ready yet. The PR request is into the multichain branch, and the same tests are failing :)

alexcos20:
right, I missed that :)

PR #578, opened 2022-12-05
Document signatures in API.md

### LINKED ISSUE #547: Document signature message
All over API.md, there is a required parameter called signature.
And described like:
```
signature: String object containg user signature (signed message)
```

Task:
 - document the message that this signature is applied to 
 (in some cases is did, or did + none, or etc..)

mariacarmina:
Towards #547 .

Changes proposed in this PR:

- added details for signature params

### DIFF
--- a/API.md
+++ b/API.md
@@ -70,16 +70,16 @@ Parameters
 
 ```
     decrypterAddress: String object containing the address of the decrypter (required)
-    chainId: the chain id of the network the document is on (required)
-    transactionId: the transaction id of the encrypted document (optional)
-    dataNftAddress: the address of the data nft (optional)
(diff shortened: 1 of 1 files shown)


API.md
```diff
@@ -634,10 +634,10 @@ Allows the user to delete an existing auth token before it naturally expires.
 Parameters
 
 ```
-    address: String object containing consumer's address (optional)
+    address: String object containing consumer's checksummed address (optional)
     nonce: Integer, Nonce (required)
     signature: String object containg user signature (signed message)
-      The signature is based on hashing the following parameters:
+      The signature is based on hashing the string concatation consisting of:
```
calina-c, line 640: "concatenation" is spelled wrongly everywhere.

PR #579, opened 2022-12-05
Fix #633: Problem: cryptic Provider log messages 

### LINKED ISSUE #633: Problem: cryptic Provider log messages
Problem: Provider has errors. When we go to debug, its logging is super cryptic which hurts ability to iterate.

Solution: add more debug code into Provider. So that when there is an issue, then it's way easier to debug.

alexcos20:
Fix #633

Problem: Provider has errors. When we go to debug, its logging is super cryptic which hurts ability to iterate.

Solution: add more debug code into Provider. So that when there is an issue, then it's way easier to debug.

Changes proposed in this PR:
- adds more debug to logs


ocean_provider/file_types/file_types_factory.py
```diff
@@ -59,8 +59,13 @@ def validate_and_create(file_obj) -> Tuple[bool, Any]:
                     userdata=file_obj.get("userdata"),
                 )
             else:
+                logger.debug(f"Unsupported type {file_obj}")
```
calina-c, line 62: shouldn't it be file_obj["type"]? I think another approach to this (and the other addition you made in this file) would be to log the message upon return, if the first member of the tuple is False. Otherwise, please assign the message variable first and use it in both debugging and the return statement.

calina-c (changes requested):

calina-c:
Status right now: we have discussed this during the weekly calls repeatedly. We are waiting for Alex to come back from holiday to determine whether this stays or goes. In case it stays, we will create an issue and link it as needed.

ocean_provider/file_types/file_types.py
```diff
@@ -132,6 +133,8 @@ def __init__(
     def validate_dict(self) -> Tuple[bool, Any]:
         if not self.url:
             return False, "missing graphql endpoint"
+        if not self.query:
+            return False, "missing graphql query"
```
calina-c, line 137: can you please add this to the possible error documentation here: https://github.com/oceanprotocol/provider/blob/main/ocean_provider/routes/README.md#file-structure-errors

calina-c: added it myself to spare you some work :)

calina-c (approved):

PR #590, opened 2023-01-06
Increase timeout to wait more for assets from aqua.

### LINKED ISSUE #589: Wait more for assets
After aquarius v5.0 update, the tests related to asset creation fail due to short timeout.

Fix: increase the timeout and sleep time from helpers function.

mariacarmina:
Fixes #589  .

Changes proposed in this PR:

- increased timeout and waiting time to retrieve assets from aquarius.
- modified time for nonces in test helpers.

Noticed the error [here](https://github.com/oceanprotocol/provider/actions/runs/3854634123/jobs/6568831886#step:9:29128)

### DIFF
--- a/tests/helpers/compute_helpers.py
+++ b/tests/helpers/compute_helpers.py
@@ -230,4 +230,4 @@ def get_compute_result(client, endpoint, params, raw_response=False):
 def get_future_valid_until(short=False):
     # return a timestamp for one hour in the future or 30s in the future if short
     time_diff = timedelta(hours=1) if not short else timedelta(seconds=30)
-    return int((datetime.utcnow() + time_diff).timestamp())
+    return int((datetime.now() + time_diff).timestamp())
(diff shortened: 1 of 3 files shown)


mariacarmina:
From the previous [run](https://github.com/oceanprotocol/provider/actions/runs/3855461075/jobs/6570535896#step:9:1033), I noticed that those 400 status codes come from `Invalid timestamp` message due to the fact that `timedelta` was set for 1 hour, instead of more or maybe the UTC time didn't work on my machine.
```python
-> expiration = int((datetime.utcnow() + timedelta(hours=1)).timestamp())
(Pdb) expiration
(Pdb) datetime.utcnow()
datetime.datetime(2023, 1, 6, 16, 34, 10, 3551)
```
For UTC time, it was 16:34 o'clock plus 1 hour delta it will be 17:34, still not enough, because on my machine was 18:34, so I tried to create an auth token with the expiration in the past that leads to `Invalid timestamp` error. That's why I modified the test helpers.

calina-c:
I don't think this is needed anymore @mariacarmina. Since the barge delay was solved, the original timeout works.

PR #605, opened 2023-02-27
Remove requests session.

calina-c:
Fixes # .

Calina: not sure if this is needed.

Background: we have some extra logic for request-sessions which is not needed. So, not sure if _this_ issue is needed here.

The first priority is to fix the custom C2D environments.

This might help that, but we need to run and see.


mariacarmina:
If we start to use `requests` package, we also need to handle retry mechanism for those requests as well. `request_session` offers support for that, see #607

trentmc:
Please link this to a github issue @calina-c

alexcos20:
let's try https://github.com/oceanprotocol/provider/pull/607 first

PR #610, opened 2023-03-06
Debug C2d connection (towards #606)

### LINKED ISSUE #606: Known issue: compute environments disappear in CI
When running the CI, the list of c2d environments disappears for no apparent reason. This affects provider and ocean.py CIs, effectively blocking further development.

We thought it was temporarily fixed by waiting through a retry mechanism here: https://github.com/oceanprotocol/provider/pull/603 but it seems that fix is not enough.

mariacarmina:
Fixes #606.


mariacarmina:
From all the investigations made so far, provider tests related to compute to data will not fail on a self-hosted runner.

**Unsuccessful** debug approaches:

- modified request headers (added `User-Agent` & `Allow-Origins`);
- replaced `request_session` with plain `requests` lib;
- created a test image of operator-service with corrected headers;
- modified OS image on the runner (tried different versions of ubuntu);
- updated python version to python3.9, 3.10;
- increased number of retries;
- replaced `urljoin` from `urllib` with strings append;
- debugged url string to search for garbage characters;
- check firewall (did not have permissions obviously on a standard github runner);
- checked C2D deployments which were fine;
- checked if listened on port 31000 and ping connection and they seem to work as expected.

**Solution**
- created a self-hosted runner with firewall disabled and the tests passed without retrying and faster

compose-env-values
```diff
@@ -5,9 +5,9 @@ PARITY_KEYFILE1=/ocean-provider/tests/resources/consumer_key_file.json
 PROVIDER_ADDRESS=068ed00cf0441e4829d9784fcbe7b9e26d4bd8d0
 
 NETWORK_URL=wss://rinkeby.infura.io/ws/v3/357f2fe737db4304bd2f7285c5602d0d
-AQUARIUS_URL=https://aquarius.marketplace.dev-ocean.com
```
calina-c, line 8: @mariacarmina  do we still need these changes? why?

ocean_provider/requests_session.py
```diff
@@ -13,7 +13,7 @@ def get_requests_session() -> Session:
     :return: requests session
     """
     session = Session()
-    retries = Retry(total=5, backoff_factor=1, status_forcelist=[502, 503, 504])
```
calina-c, line 16: how about these ones?

mariacarmina: we should keep retries in case of anything.

compose-env-values
mariacarmina: We don't need them I just saw that the URLs were deprecated.

calina-c (approved):

mariacarmina:
UPDATE:
I've tested certain files to see which one is taking much more time and it seems that `test_initialize` needs to be optimised, because it took an hour to be tested.

```console
test_download.py ..................... 3m24s
test_fileinfo.py......................4m16s
test_encryption.py....................5s
test_initialize.......................1h3m11
```

start_provider.sh
```diff
@@ -2,7 +2,7 @@
 export PROVIDER_PRIVATE_KEY=0xfd5c1ccea015b6d663618850824154a3b3fb2882c46cefb05b9a93fea8c3d215
 export TEST_PRIVATE_KEY1=0xef4b441145c1d0f3b4bc6d61d29f5c6e502359481152f869247c7a4244d45209
 export TEST_PRIVATE_KEY2=0x5d75837394b078ce97bc289fa8d75e21000573520bfa7784a9d28ccaae602bf8
-export OPERATOR_SERVICE_URL=https://c2d-dev.operator.oceanprotocol.com/
+export OPERATOR_SERVICE_URL=http://172.15.0.13:31000
```
calina-c, line 5: why do we need this? Are we sure this is not used in production, like it was with start_aquarius.sh?

calina-c (changes requested):
Please make sure you also search for TODOs in the code or comments containing the issue number (606, 610)

start_provider.sh
mariacarmina: I have updated the URLs. This is a script for development environment and that is the local URL for operator service, not
`https://c2d-dev.operator.oceanprotocol.com/`.

tests/test_graphql.py
```diff
@@ -137,7 +137,10 @@ def test_download_graphql_asset_with_userdata(
...
     assert response.status_code == 200, f"{response.data}"
     reply = json.loads(response.data)
+    assert len(reply["errors"]) == 1
+    # Make sure the subgraph is deployed. Response has 200 OK status code from subgraph
+    # Due to lack of CI resources, assert the specific error
     assert (
-        len(reply["data"]["nfts"]) == 1
-    )  # make sure our parametrized query works, otherwise we will get a lot of nfts
-    assert reply["data"]["nfts"][0]["id"] == asset.nftAddress.lower()
+        reply["errors"][0]["message"]
```
calina-c, line 144: wait, why? We should be able to retrieve assets from subgraph. Why are we asserting that subgraph is not deployed?

calina-c (changes requested):

tests/test_graphql.py
mariacarmina: This is the only test that fails due to the fact that I have added in the flow "--skip-subgraph-deploy". that's why I wanted to assert that specific order. The CI does not support c2d, subgraph, RBAC all in one workflow. To reduce the space, I chose to skip the subgraph deployment. If I completely remove subgraph from the workflow, multiple tests from graphql fail. So, I have decided to keep subgraph and assert the specific error.

mariacarmina: I can delete the assert of the message (if needed), but I just wanted to highlight the error first just so we can be aware of it when we skip subgraph deployment.

cc @alexcos20

alexcos20 (approved):
lgtm, we can merge it to unblock

tests/test_graphql.py
```diff
@@ -79,13 +79,13 @@ def test_download_graphql_asset_with_userdata(
     unencrypted_files_list = [
         {
             "type": "graphql",
-            "url": "http://172.15.0.15:8000/subgraphs/name/oceanprotocol/ocean-subgraph",
+            "url": "http://172.15.0.15:8030/graphql",
             "query": """
-                    query nfts($nftAddress: String){
```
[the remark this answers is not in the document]
mariacarmina: done

tests/test_graphql.py
mariacarmina: done by adding a new subgraph query

PR #615, opened 2023-03-13
Skip if setup fails for environments.

calina-c:
Fixes # .

Changes proposed in this PR:

- 
- 
-

### DIFF
--- a/ocean_provider/utils/compute_environments.py
+++ b/ocean_provider/utils/compute_environments.py
@@ -21,20 +21,14 @@ def get_c2d_environments() -> List:
     standard_headers = {"Content-type": "application/json", "Connection": "close"}
     web3 = get_web3()
     params = {"chainId": web3.eth.chain_id}
-    response = None
-    retries = 2
-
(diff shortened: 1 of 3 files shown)


ocean_provider/utils/compute_environments.py
```diff
@@ -20,9 +21,22 @@ def get_c2d_environments() -> List:
     standard_headers = {"Content-type": "application/json", "Connection": "close"}
     web3 = get_web3()
     params = {"chainId": web3.eth.chain_id}
-    response = requests_session.get(
-        get_compute_environments_endpoint(), headers=standard_headers, params=params
-    )
+    response = None
+    retries = 2
+
+    while retries != 0:
```
mariacarmina, line 27: if you use request_session, which you do, you don't need retry mechanism here because the HTTPAdapter has max_retries set. So keep it how it was initially please.

mariacarmina (changes requested):

mariacarmina (approved):

PR #618, opened 2023-03-14
Adds event searching.

### LINKED ISSUE #617: Incorrect event validation for download/compute

For instance:
```
        tx_receipt, errors=DISCARD
    )

    provider_fee_order_log = (
    )
```

This is also valid for:
```
````

This needs to be fixed asap.

Why:

calina-c:
Closes #617.

### DIFF
--- a/ocean_provider/utils/datatoken.py
+++ b/ocean_provider/utils/datatoken.py
@@ -69,9 +69,12 @@ def verify_order_tx(
         tx_receipt, errors=DISCARD
     )
 
-    provider_fee_order_log = (
-        provider_fee_event_logs[0] if provider_fee_event_logs else None
-    )
+    provider_fee_order_log = None
+    for log in provider_fee_event_logs:
+        if log.event == "ProviderFee":
+            provider_fee_order_log = log
+            break
+
     if not provider_fee_order_log:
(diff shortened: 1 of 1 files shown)


alexcos20 (commented):
~~they are equivalent in terms of results, and existing version is more concise
So we don't need this~~
if multiple events exists in a tx, this will return only the last one.  This is wrong

calina-c:
> ~they are equivalent in terms of results, and existing version is more concise So we don't need this~ if multiple events exists in a tx, this will return only the last one. This is wrong

Yes, that's why I'm proposing it as a temporary solution. It is better than the previous version, where the first event, regardless of type, was checked. This one at least targets the correct type. We can discuss a better filtering strategy when you come back from holiday.

alexcos20:
`provider_fee_event_logs = datatoken_contract.events.ProviderFee().processReceipt(
        tx_receipt, errors=DISCARD
    )`
This returns all events of type 'ProviderFee' from tx_receipt, so we have them all, no matter if it's only one or one hundred.
It will not store in provider_fee_event_logs  events of type 'OrderStarted' for instance.

Taking either the first or the last is wrong. 
We need to process them all and see which one is the one we are looking for.

PR #623, opened 2023-03-24
restrict fileinfo for url to only GET method

soonhuat:
Fixes # 621.

Changes proposed in this PR:

- validation to restrict url type fileinfo to GET method only

### DIFF
--- a/ocean_provider/file_types/file_types.py
+++ b/ocean_provider/file_types/file_types.py
@@ -31,7 +31,7 @@ class UrlFile(EndUrlType, FilesType):
         if not self.url:
             return False, "malformed service files, missing required keys."
 
-        if self.method not in ["get", "post"]:
+        if self.method not in ["get"]:
             return False, f"Unsafe method {self.method}."
 
         return True, self


soonhuat:
@calina-c looks like previously already has checking on fileinfo method based on file types, I removed POST for **url file types**
so POST should still allowed for **GraphQL file types**
please abandon if this doesn't make sense
@mihaisc

calina-c:
@soonhuat This change disallows checking for "post" file types, which is a functionality, as far as I remember. When we discussed this, we were looking at two entrypoints:
- the request done by a user/agent towards provider. I promised to look into it and it is indeed restricted to POST, but that is by design since we need to have a payload containing either "did"+"serviceId" or "type"+"hash"+file specifics (i.e. url or arweave transaction)
- the request done by provider towards the resource, which can be either "get" or "post" as defined by the user when publishing the asset. E.g. if the file is only available via POST download, we should allow that.

As far as I can tell, all is normal behaviour. Can you give me an example where this is troublesome?

soonhuat:
> checking for "post" file types

@calina-c 
if strictly follow RESTful API standard then should be only GET, on the hand POST could post risk of server side forgery, maybe the chances of happening are not high
is ok if current behaviour is per ocean generic use case, then this PR can be abandon.

calina-c:
Closing this, since this is the intended flow.

PR #630, opened 2023-03-30
WIP: Process all receipts partially.

### LINKED ISSUE #617: Incorrect event validation for download/compute

calina-c:
Closes #617


calina-c:
Superseded by #632

PR #632, opened 2023-04-04
Fix Incorrect event validation

### LINKED ISSUE #617: Incorrect event validation for download/compute

alexcos20:
Fixes #617 

Changes proposed in this PR:

- 
- 
-

### DIFF
--- a/ocean_provider/utils/datatoken.py
+++ b/ocean_provider/utils/datatoken.py
@@ -65,20 +65,28 @@ def verify_order_tx(
 
     # check provider fees
     datatoken_contract = get_datatoken_contract(web3, datatoken_address)
+    provider_fee_order_log = None
     provider_fee_event_logs = datatoken_contract.events.ProviderFee().processReceipt(
         tx_receipt, errors=DISCARD
     )
+    # search in all provider_fee events until we have a match. if not, we don't have a valid event
+    # also, make sure that somebody is not spoofing provider fee event from another datatoken
+    for provider_fees_logs in provider_fee_event_logs:
+        try:
+            provider_data = json.loads(provider_fees_logs.args.providerData)
+            if (
+                provider_data["dt"].lower() == datatoken_address.lower()
+                and provider_data["id"].lower() == service.id.lower()
+            ):
+                provider_fee_order_log = provider_fees_logs
 
-    provider_fee_order_log = (
-        provider_fee_event_logs[0] if provider_fee_event_logs else None
-    )
+        except:
+            # silent pass, means json formatting errors
+            pass
+
     if not provider_fee_order_log:
         raise AssertionError(
             f"Cannot find the event for the provider fee in tx id {tx_id}."
-        )
-    if len(provider_fee_event_logs) > 1:
-        raise AssertionError(
-            f"Multiple order events in the same transaction !!! {provider_fee_order_log}"
         )
 
     provider_initialize_timestamp = 0
@@ -176,15 +184,15 @@ def verify_order_tx(
     except Exception as e:
         logger.error(e)
     logger.debug(f"Got events log when searching for OrderStarted : {event_logs}")
-    order_log = event_logs[0] if event_logs else None
+    order_log = None
+    # search in all startOrder events until we have a match. if not, we don't have a valid event
+    for log in event_logs:
+        if log.args.serviceIndex == service.index:
+            order_log = log
 
     if not order_log:
         raise AssertionError(
             f"Cannot find the event for the order transaction with tx id {tx_id}."
-        )
-    if len(event_logs) > 1:
-        raise AssertionError(
-            f"Multiple order events in the same transaction !!! {event_logs}"
         )
 
     if order_log.args.serviceIndex != service.index:
--- a/tests/test_download.py
+++ b/tests/test_download.py
@@ -3,6 +3,7 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 import copy
+import logging
 import time
 from datetime import datetime
 from unittest.mock import patch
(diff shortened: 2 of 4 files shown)


ocean_provider/utils/datatoken.py
```diff
@@ -65,21 +65,29 @@ def verify_order_tx(
...
+    for provider_fees_logs in provider_fee_event_logs:
+        try:
+            provider_data = json.loads(provider_fees_logs.args.providerData)
+            if (
+                provider_data["dt"].lower() == datatoken_address.lower()
+                and provider_data["id"].lower() == service.id.lower()
+            ):
+                provider_fee_order_log = provider_fees_logs
+
+        except:
```
calina-c, line 83: please catch the json encoding error here separately and raise the appropriate AssertionError

calina-c: also, can you please wrap just the problematic function using try/except? The second condition should be separate

ocean_provider/utils/datatoken.py
```diff
@@ -176,16 +184,16 @@ def verify_order_tx(
     except Exception as e:
         logger.error(e)
     logger.debug(f"Got events log when searching for OrderStarted : {event_logs}")
-    order_log = event_logs[0] if event_logs else None
+    order_log = None
```
calina-c, line 187: what about the order reused log? Shouldn't it also be handled similarly?

calina-c (changes requested):

tests/test_helpers.py
```diff
@@ -500,3 +531,37 @@ def build_custom_services(
 def get_first_service_by_type(asset, service_type: ServiceType) -> Service:
     """Return the first Service with the given ServiceType."""
     return next((service for service in asset.services if service.type == service_type))
+
+
+def get_service_by_index(asset, index: int) -> Service:
+    """Return the first Service with the given ServiceType."""
+    i = 0
+    for service in asset.services:
```
calina-c, line 539: please use the `next` syntax, like in ocean.py

ocean_provider/utils/datatoken.py
```diff
@@ -65,21 +65,29 @@ def verify_order_tx(
...
     provider_fee_event_logs = datatoken_contract.events.ProviderFee().processReceipt(
         tx_receipt, errors=DISCARD
     )
+    # search in all provider_fee events until we have a match. if not, we don't have a valid event
+    # also, make sure that somebody is not spoofing provider fee event from another datatoken
+    for provider_fees_logs in provider_fee_event_logs:
+        try:
+            provider_data = json.loads(provider_fees_logs.args.providerData)
+            if (
+                provider_data["dt"].lower() == datatoken_address.lower()
```
calina-c, line 78: also, can you please check this key exists? It won't exist on older requests

alexcos20: see answer above

ocean_provider/utils/datatoken.py
alexcos20: why ?
the logic is the following:
 - iterate over existing events
 - if event has a malformed json (or none at all), or if dt key does not exists or if id key does not exists or if dt address is not the expecting one or ....  , then this is not what we are looking for.

I don't care why it fails, that why there is only one try/catch..  as long as something failed, this event is not what we are looking for

ocean_provider/utils/datatoken.py
alexcos20: no, because they are already filtered using data token address

tests/test_download.py
```diff
@@ -96,6 +104,87 @@ def test_download_service(
         assert response.status_code == 400, f"{response.data}"
 
 
+@pytest.mark.integration
+@pytest.mark.parametrize(
+    "userdata,erc20_enterprise",
```
calina-c, line 109: I'm almost ready to approve, in the meantime I noticed all the parametrisations are copied from the single-order test. Do we need them in both places? Are they properly used in both tests? Maybe we can keep them in just one place, since there's no point in testing the same functionality multiple times, in 8 different tests.

tests/test_download.py
```diff
@@ -96,6 +104,87 @@ def test_download_service(
...
+        web3, service1.datatoken_address, consumer_wallet.address, publisher_wallet
+    )
+    approve_tokens(
+        web3, service1.datatoken_address, nft_factory_address, 1, consumer_wallet
+    )
+    # create an asset with two services
+    asset2 = get_registered_asset(
+        publisher_wallet, erc20_enterprise=erc20_enterprise, no_of_services=2
+    )
+    service2 = get_first_service_by_type(asset2, ServiceType.ACCESS)
```
calina-c, line 128: could you please move hte repeated instructions to a helper function in order to improve readability? I'm talking about get_registered_assrt, get_first_service_by_type, minting and approving, which seems to be copied 3 times.

tests/test_download.py
```diff
@@ -96,6 +104,87 @@ def test_download_service(
...
+                    0,
+                ),
+            ),
+            (
+                service2.datatoken_address,
+                consumer_wallet.address,
+                service2.index,
+                get_provider_fees(asset2, service2, consumer_wallet.address, 0),
+                (
+                    "0x0000000000000000000000000000000000000000",
```
calina-c, line 163: I think you can use ZERO_ADDRESS here (and in other places) instead

calina-c (changes requested):

tests/test_download.py
alexcos20: there is no ZERO_ADDRESS defined, so let's create another task for this (very low prio)

tests/test_download.py
alexcos20: done in https://github.com/oceanprotocol/provider/pull/632/commits/f5c7edbc4bf29880742dcf7375b55a58a0de269e

tests/test_download.py
alexcos20: agreed

calina-c (approved):

PR #638, opened 2023-04-10
Reinstate old_signature check.

### LINKED ISSUE #637: Restore old_signature check
Pending TODO from here: https://github.com/oceanprotocol/provider/blob/459cb4b35b77d08e828ea3a88f9aa6eca72e054d/ocean_provider/utils/accounts.py#L47

calina-c:
Closes #637. This is still not clear to me. Why do we have this check? We are the ones realigning the signature. What could we expect differently?


calina-c:
Superseded by #639

PR #646, opened 2023-06-08
Split runners

### LINKED ISSUE #647: C2d tests fail on CI
**Describe the bug**
C2d tests fail on CI when retrieving C2D envs with the following error: "Connection refused".

**Expected behavior**
Tests pass.

mariacarmina:
Fixes #647  .

Changes proposed in this PR:

- split workflows, one for c2d and one for the other tests.
- updated compute test to omit RBAC
- added markers for compute, algo_validation and provider_fees tests which contain C2D functions
- update pytest.ini


.github/workflows/c2d.yml
```diff
@@ -0,0 +1,67 @@
...
+      - name: Test with pytest
+        run: |
+          pytest tests/test_compute.py
+          pytest ocean_provider/utils/test/test_compute.py
+          pytest ocean_provider/utils/test/test_provider_fees.py
+          pytest ocean_provider/validation/test/test_algo_validation.py
+      - name:  docker logs
+        run: docker logs ocean_aquarius_1 && docker logs ocean_provider_1 && docker logs ocean_provider2_1 && docker logs ocean_ipfs_1
+        if: ${{ failure() }}
+      - name: Publish code coverage
```
calina-c, line 64: so now we are only reporting coverage from the c2d part? That is not correct. Coverage reports and badges are very important. What coverage do we get just by running c2d tests? That reflects poorly on our code.

calina-c (changes requested):

tests/test_fileinfo.py
```diff
@@ -100,10 +102,19 @@ def test_checksums(client):
         assert file_info["checksumType"] == "sha256"
 
     # big file, we should not have a checksum
+    response = requests.get(
+        "https://raw.githubusercontent.com/t-davidson/hate-speech-and-offensive-language/master/data/labeled_data.csv"
```
alexcos20, line 106: can we use datasets from our own github? we can create a new repo for this if needed

alexcos20 (changes requested):

tests/test_fileinfo.py
mariacarmina: I think creating a separate repo for datasets is a great idea as we have for c2d algo.
I'll replace the URL with what we have already.

.github/workflows/c2d.yml
mariacarmina: Yes, it is a possible way to combine the results for coverage report https://docs.knapsackpro.com/2019/how-to-merge-codeclimate-reports-for-parallel-jobs-ci-nodes. I'll try that approach.
cc @alexcos20

mariacarmina:
Discovered the problem by debugging with SSH inside the runner, found out that the operator service image is too large and exceeds the disk space capacity. Need to do a resize to those Docker images in a separate issue.

mariacarmina:
Closing this since the problem needs to be solved on the majority of the repos.

PR #649, opened 2023-07-03
Refactor user nonce logic.

### LINKED ISSUE #648: Unable to purchase computed data with mumbai testnet due to Invalid signature error
**Describe the bug**
Getting an 400 bad request when trying to purchase an compute only data set, published from the different account , from different browser. Using mumbai testnet. Algorithms , dataset refered from here ocean.py tutorial. 
This issue is reported from [Discord](https://discord.com/channels/612953348487905282/1120325871732019201). 

**To Reproduce**
1. Go to ocean.py/READMEs/c2d-flow.md
2. Publish the dataset and the algorithm assets
3. Open Ocean Market in the browser and search for published assets.
4. Try to purchase service and see the error

**Expected behavior**
To purchase the services without problems.

**Screenshots**
image

```console
ocean_lib.exceptions.DataProviderException: Failed to get a response for request:
computeStartEndpoint=https://v4.provider.mumbai.oceanprotocol.com/api/services/compute, payload={'dataset':
{'documentId': 'did:op:443a22ebed56b9c846dafb751e4cef47b1869d77d1535a9865175b27b3c7adcd',
'serviceId': '1', 'transferTxId': '0x56024d9c2c1d2d3b3021f21dcc5cf6c544914d8fa6ddb371be7b2232f5f99918'},
'environment': 'compute-v4-engine-dev', 'algorithm': {'documentId':
'did:op:09aa3f5235ed3170baa2eb5cacb8c64068ac444f29209cf792637eea98fde198', 'serviceId': '0', 'transferTxId':
'0x17dc3e5510e5786fc2fec1aa988017de1926cee8909d085f2f91df61471f6096'}, 'signature':
'0xa657ac434840374f19dbd4c9bb60d1683dcab17663c24bcf36069ede3d3804e475430178b9ca44a62c08e653280f31a8f738e479ca9d88c016960c492f93d7ff01', 'nonce': '1687177454529.354', 'consumerAddress':
'0x0b600C9d669241453A449f766fF49b5CC50eDD54', 'additionalInputs': []},
response is b'<html>\r\n<head><title>502 Bad Gateway</title></head>\r\n<body>\r\n<center><h1>502 Bad Gateway</h1></center>\r\n<hr><center>nginx</center>\r\n</body>\r\n</html>\r\n'
```

mariacarmina:
Fixes #648  .

Changes proposed in this PR:

- solve Invalid signature error
- fix bug in nonce endpoint
- convert & store provider nonces into integers.


ocean_provider/routes/compute.py
```diff
@@ -246,11 +247,13 @@ def computeDelete():
     data = get_request_data(request)
     logger.info(f"computeDelete called. arguments = {data}")
 
-    body = process_compute_request(data)
-    response = requests_session.delete(
-        get_compute_endpoint(), params=body, headers=standard_headers
-    )
-    update_nonce(body["owner"], data.get("nonce"))
+    lock = Lock()
+    with lock:
```
alexcos20, line 251: we are not using threads, so this is useless

bogdanfazakas:
@mariacarmina i think there is a use case that is not well handled, if the address sent is a fresh one the provider returns a null response instead of  creating a new entrance for the new address and return 0 maybe

mariacarmina:
> @mariacarmina i think there is a use case that is not well handled, if the address sent is a fresh one the provider returns a null response instead of creating a new entrance for the new address and return 0 maybe

I suggest to create a new entry an return the new nonce value instead of 0 or null.

bogdanfazakas:
> > @mariacarmina i think there is a use case that is not well handled, if the address sent is a fresh one the provider returns a null response instead of creating a new entrance for the new address and return 0 maybe
> 
> I suggest to create a new entry an return the new nonce value instead of 0 or null.

yep agreed with the new nonce value, 0 was just a better alternative for null

mariacarmina:
Tested on barge & mumbai a compute flow (end to end) and it worked as expected, without occurring `InvalidSignatureError `(I could start the C2D job, get the status & results of the algorithm, get the C2D envs successfully).
To mention that I tested these modifications on a separate VM (hosted in US) which has better hardware resources.

ocean_provider/utils/compute.py
```diff
@@ -46,7 +47,7 @@ def process_compute_request(data):
 
 
 def sign_for_compute(wallet, owner, job_id=None):
-    nonce = datetime.now(timezone.utc).timestamp()
+    nonce = datetime.now(timezone.utc).timestamp() * 1000
```
alexcos20, line 50: if we handle this in a new issue, then lgtm

alexcos20 (approved):

PR #655, opened 2023-08-01
Rejected HEAD & OPTIONS requests where nonce is updated in the DB. Add filename to fileinfo endpoint.

### LINKED ISSUE #656: Rejected HEAD & OPTIONS requests where nonce is updated in the DB. Add filename to fileinfo endpoint.
Reject HEAD & OPTIONS requests where nonce is updated in the DB. 
Add filename to fileinfo endpoint to be easier for frontend to fetch the filename without sending HEAD requests.

mariacarmina:
Fixes #656  .

Changes proposed in this PR:

- rejected HEAD & OPTIONS requests where nonce is updated in the DB
- add filename to fileinfo endpoint
- tests

### DIFF
--- a/ocean_provider/file_types/definitions.py
+++ b/ocean_provider/file_types/definitions.py
@@ -88,12 +88,9 @@ class EndUrlType:
                     filename = urlparse(files_url).path.split("/")[-1]
                     try:
                         if not self._validate_filename(filename):
-                            logger.error(
-                                "Invalid file name format. It was not possible to get the file name."
-                            )
-                            return False, {
-                                "err_msg": "Invalid file name format. It was not possible to get the file name."
-                            }
+                            msg = "Invalid file name format. It was not possible to get the file name."
+                            logger.error(msg)
+                            return False, {"error": msg}
 
                         file_name = filename
                     except Exception as e:
--- a/ocean_provider/file_types/definitions.py
+++ b/ocean_provider/file_types/definitions.py
@@ -94,10 +94,9 @@ class EndUrlType:
 
                         file_name = filename
                     except Exception as e:
-                        logger.warning(f"It was not possible to get the file name. {e}")
-                        return False, {
-                            "err_msg": f"It was not possible to get the file name. {e}"
-                        }
+                        msg = f"It was not possible to get the file name. {e}"
+                        logger.warning(msg)
+                        return False, {"error": msg}
 
                 if not content_length and content_range:
                     # sometimes servers send content-range instead
--- a/ocean_provider/file_types/definitions.py
+++ b/ocean_provider/file_types/definitions.py
@@ -4,9 +4,11 @@ import json
 import logging
 import mimetypes
 import os
+import re
 from abc import abstractmethod
 from cgi import parse_header
 from typing import Protocol, Tuple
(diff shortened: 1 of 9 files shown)


mariacarmina:
@alexcos20 Let me know if you want to create a decorator function instead of having those conditions e.g. `if request.method == 'HEAD' or request.method == 'OPTIONS'`, this is more like a cosmetic change.

ocean_provider/file_types/definitions.py
```diff
@@ -78,6 +82,25 @@ def check_details(self, with_checksum=False):
...
                 content_length = headers.get("Content-Length")
                 content_range = headers.get("Content-Range")
+                file_name = None
+
+                if files_url:
+                    filename = urlparse(files_url).path.split("/")[-1]
+                    try:
+                        if not self._validate_filename(filename):
+                            logger.error(
+                                "Invalid file name format. It was not possible to get the file name."
```
calina-c, line 92: to avoid duplicating this string when changing it, set it up as a variable e.g. msg="Invalid file format...", then logger.error(msg) and return False, {"err_msg": msg}. Also, can you please make it "error" instead of "err_msg"? It should look like the other errors to allow frontend to make their logic based on error messages. And please document it in the readme where we explain all the error formats.

ocean_provider/routes/compute.py
```diff
@@ -240,9 +240,14 @@ def computeDelete():
         description: One or more of the required attributes are missing or invalid.
       401:
         description: Invalid asset data.
+      405:
+        description: Method rejected.
       503:
         description: Service Unavailable
     """
+    if request.method == "HEAD" or request.method == "OPTIONS":
```
calina-c, line 248: I remember I grouped these in a list called "safe methods" or something similar. You can use something like if request.method in safe_methods: ...

And then if we import these "safe methods" from another file you won't have to apply the change multiple times in the future, e.g. if a new restriction is made.

ocean_provider/routes/decrypt.py
```diff
@@ -90,9 +90,14 @@ def decrypt():
         description: decrypted document
       400:
         description: One or more of the required attributes are missing or invalid.
+      405:
```
calina-c, line 93: please document the new status codes as well, in the readmes.

calina-c (changes requested):

ocean_provider/routes/compute.py
mariacarmina: There are no `safe_methods` or certain lists for it but I can do something like:

```
if request.method in ['HEAD', 'OPTIONS']
```

calina-c: yeah, I think you can create the list in some constants file. I would also make sure to use upper for request method.

ocean_provider/routes/decrypt.py
calina-c: do not forget about the error readme :) https://github.com/oceanprotocol/provider/blob/main/ocean_provider/routes/README.md

mariacarmina: I have put them there as well :)

calina-c (approved):

PR #667, opened 2023-10-03
Enforce algo credentials check

### LINKED ISSUE #657: Credetials is not enforced, at least in the case of an algorithm
**Describe the bug**

After publishing an algorithm and restricting its use using the credentials feature:

{
  "allow": [
      {
          "type": "address",
          "values": ["0x620...6822"]
      }
  ],
  "deny": []
}

The credentials are not enforced and the algorithm can be run by other users different from the allowed one.

**To Reproduce**
Tested in the context of the Pontus-X project which uses as default provider https://provider.v4.genx.delta-dao.com, which is on version 1.3.12.

The algorithm was published using the Nautilus library, which allows to set the credentials restriction programmatically, as it is not available from the Market front end.

Not able to test if this issue is still present in version 2.1.0, though I've been unable to find any issue related to this problem addressed from version 1.3.12.

**Expected behavior**
Just the users explicitly granted access through credentials can execute the algorithm.

mariacarmina:
Fixes #657  .

Changes proposed in this PR:

- enforce credentials check for algorithms
- tests

### DIFF
--- a/ocean_provider/validation/algo.py
+++ b/ocean_provider/validation/algo.py
@@ -14,6 +14,8 @@ from ocean_provider.utils.asset import (
     get_asset_from_metadatastore,
 )
 from ocean_provider.utils.basics import get_metadata_url, get_provider_wallet, get_web3
+from ocean_provider.utils.consumable import ConsumableCodes
+from ocean_provider.utils.credentials import AddressCredential
 from ocean_provider.utils.datatoken import (
     record_consume_request,
     validate_order,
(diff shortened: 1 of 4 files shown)


mariacarmina:
Fixed workflow

alexcos20 (approved):
lgtm

rogargon:
Great work @mariacarmina, thanks for fixing #657

mariacarmina:
> Great work @mariacarmina, thanks for fixing #657

We'll release a new Provider soon as well with the latest modifications. Thank you!

PR #669, opened 2023-10-09
Added network name in logs.

### LINKED ISSUE #644: Add network-name in each of the provider logs
As we now have the multichain provider, reading the logs is a bit difficult. 
I'd love to have a way to filter the logs for a specific network. To do this, we need to log the network name with each of the provider logs

mariacarmina:
Fixes #644  .

Changes proposed in this PR:

- add network name in logs

### DIFF
--- a/ocean_provider/utils/basics.py
+++ b/ocean_provider/utils/basics.py
@@ -20,6 +20,17 @@ from web3.middleware import geth_poa_middleware
 
 logger = logging.getLogger(__name__)
 
+NETWORK_NAME_MAP = {
+    1: "Mainnet",
+    5: "Goerli",
+    56: "Binance Smart Chain",
+    137: "Polygon",
+    246: "Energy Web",
+    1285: "Moonriver",
+    80001: "Mumbai",
+    11155111: "Sepolia",
+}
+
 
 def decode_keyed(env_key):
     try:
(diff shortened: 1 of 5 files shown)


ocean_provider/utils/basics.py
```diff
@@ -20,6 +20,18 @@
 
 logger = logging.getLogger(__name__)
 
+NETWORK_NAME_MAP = {
```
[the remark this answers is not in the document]
mariacarmina: I'll use addresses.json

ocean_provider/utils/basics.py
```diff
@@ -161,6 +164,24 @@ def get_web3_connection_provider(
...
+    address_path = Path(os.path.join(addresses.__file__, "..", "address.json"))
+
+    address_file = address_path.expanduser().resolve()
+    with open(address_file) as f:
+        addresses_json = json.load(f)
+
+    for k, v in addresses_json.items():
+        if v["chainId"] == chain_id:
+            return k
+
```
alexcos20, line 183:
```suggestion
return ("Unknown")
```

alexcos20: maybe provider works on a custom network. let's return "Unknown" if that's the case

alexcos20 (commented):
other than that, lgtm , but I'm worried about tests

alexcos20 (approved):

PR #672, opened 2023-11-16
Fix fileinfo for arweave file

mariacarmina:
Fixes # .

Changes proposed in this PR:

- Added check for file path validation inside UrlFile class.
- Fix fileinfo for arweave file.
- Updated tests.

### DIFF
--- a/ocean_provider/file_types/definitions.py
+++ b/ocean_provider/file_types/definitions.py
@@ -72,7 +72,10 @@ class EndUrlType:
                 if result:
                     status_code = result.status_code
                     headers = copy.deepcopy(result.headers)
-                    files_url = copy.deepcopy(result.url)
+                    if "arweave" in result.url:
+                        files_url = ""
+                    else:
+                        files_url = copy.deepcopy(result.url)
                     # always close requests session, see https://requests.readthedocs.io/en/latest/user/advanced/#body-content-workflow
                     result.close()
                     if status_code == 200:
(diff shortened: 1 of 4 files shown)


ocean_provider/file_types/definitions.py
```diff
@@ -72,7 +72,10 @@ def check_details(self, with_checksum=False):
                 if result:
                     status_code = result.status_code
                     headers = copy.deepcopy(result.headers)
-                    files_url = copy.deepcopy(result.url)
+                    if "arweave" in result.url:
```
alexcos20, line 75: what if arweave gateway url is  "http://gateway.ar.example" ?
honestly, I would move this to Arweave class  (overwrite check_details) and do it for every arweave file

ocean_provider/file_types/definitions.py
```diff
@@ -72,7 +72,10 @@ def check_details(self, with_checksum=False):
...
+                        files_url = ""
```
alexcos20, line 76: can we get at least Content-size ?

alexcos20 (changes requested):

ocean_provider/file_types/definitions.py
mariacarmina: I return content-length and content-type, these come from headers.

ocean_provider/file_types/definitions.py
mariacarmina: Sure, thanks for pointing out.

ocean_provider/file_types/definitions.py
mariacarmina: I assume you referred to Content-Length

alexcos20 (approved):
lgtm

PR #676, opened 2024-02-07
add more details for malformed addrs

alexcos20:
If either datatokenAddress or nftAddress are malformed in the encrypted files string

(example:
```
b'{"datatokenAddress":"0x0","nftAddress":"0x0","files":[{"type":"url","url":"https://github.com/......0Safety%20Traffic%20Accident%20Analysis%2028eaf51fb56f4fbe80cb1ac0e552d384.pdf","method":"GET"}]}'
```

,provider fails with:

```
2024-02-07 05:59:01 provider-59c9fdb476-fl7pt ocean_provider.utils.util[12] ERROR Provider polygon: Error decrypting service files <class 'ocean_provider.utils.services.Service'>: Unknown format 0x0, attempted to normalize to 0x0
2024-02-07 05:59:01 provider-59c9fdb476-fl7pt ocean_provider.routes.consume[18] ERROR Provider polygon: Unable to get dataset files
```

Why?  cause this is where it fails:   https://github.com/oceanprotocol/provider/blob/main/ocean_provider/utils/util.py#L58
Because Web3.toChecksumAddress complains about unknown format "0x0".. 

Obviously, this will fail the validation, but it will give you a clue why it failed
This PR adds more detailed logs

PS: This also updates .github/workflows/pytest.yml

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -41,24 +41,14 @@ jobs:
           pip install -r requirements_dev.txt
       - name: Delete default runner images
         run: |
-          docker image rm node:14
-          docker image rm node:14-alpine
           docker image rm node:16
           docker image rm node:16-alpine
           docker image rm node:18
           docker image rm node:18-alpine
           docker image rm node:20
-          docker image rm node:20-alpine
-          docker image rm buildpack-deps:buster
-          docker image rm buildpack-deps:bullseye
           docker image rm debian:10
           docker image rm debian:11
           docker image rm moby/buildkit:latest
-          docker image rm alpine:3.16
-          docker image rm alpine:3.17
-          docker image rm alpine:3.18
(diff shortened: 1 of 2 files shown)


.github/workflows/pytest.yml
```diff
@@ -41,24 +41,14 @@ jobs:
           pip install -r requirements_dev.txt
       - name: Delete default runner images
         run: |
-          docker image rm node:14
```
mariacarmina, line 44: Why should we delete those lines? those images are no longer pulled by barge?

alexcos20: they don't exist anymore in github runner

alexcos20: see https://github.com/oceanprotocol/provider/actions/runs/7810951839/job/21305187944

